PR #31, opened 2020-01-12
Fix prefer lowest compatibility

dereuromark:
Trying to fix
```
1) IdeHelper\Test\TestCase\Annotator\ClassAnnotatorTask\TestClassAnnotatorTaskTest::testAnnotate
TypeError: Return value of Zend\Diactoros\marshalUriFromSapi() must be an instance of Zend\Diactoros\Uri, instance of Laminas\Diactoros\Uri returned
/home/travis/build/dereuromark/cakephp-ide-helper/vendor/laminas/laminas-diactoros/src/functions/marshal_uri_from_sapi.legacy.php:20
```
https://travis-ci.org/dereuromark/cakephp-ide-helper/jobs/635958965?utm_medium=notification&utm_source=github_status

But I am not sure if this actually fixes things.
Tests are still failing https://github.com/dereuromark/cakephp-ide-helper/pull/168

I would expect prefer-lowest to not have aliasing issues as this compromises those important travis checks.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
     },
     "require": {
         "php": "^7.1",
-        "laminas/laminas-zendframework-bridge": "^1.0",
+        "laminas/laminas-zendframework-bridge": "^1.0.1",
         "psr/http-factory": "^1.0",
         "psr/http-message": "^1.0"
     },


geerteltink:
To fix your issues, you might want to look at your cakephp requirement. In your composer file you require 3.7: https://github.com/dereuromark/cakephp-ide-helper/blob/master/composer.json#L26
cakephp switched to laminas in 4.0.2

I'm guessing the exceptions are caused because of typehinting in cakephp <4.0.2

dereuromark:
You are on the wrong branch: cake4 ( https://github.com/dereuromark/cakephp-ide-helper/pull/168/files )
But yeah, only allowing latest patch works for now. I guess this is the only way to solve it for the time being on our end.

geerteltink:
Ah ok, I was checking your master branch.

Xerkus:
@dereuromark problem is with a bug that was fixed in patched tags eg 2.2.0p1
https://github.com/laminas/laminas-diactoros/blob/74142df5008e53900f87b68f2c64cf7e6090e606/src/functions/create_uploaded_file.legacy.php#L18-L21

Prefer lowest installs  2.2.0 that does not have that fix.

dereuromark:
Raising the min then works for now. THX.

Xerkus:
@dereuromark raising min does not fix it really. Bug affects all migrated tags. Normally composer would select patched tag but prefer lowest makes it prefer base tag instead.

Xerkus:
I found no way to fix the `--prefer-lowest` use case here. However impact is minimal as this is an edge case encountered in the CI environment. It will be left as won't fix.

@dereuromark There is a way to fix it on your side by declaring conflicts with laminas-diactoros by listing each tag within the supported range that have corresponding patch tag, eg `2.2.0p1`:
```json
    "conflict": {
        " laminas/laminas-diactoros": "2.1.0 || 2.1.1 || 2.1.5 || 2.2.0 || 2.2.1"
    }
```
This will allow only patched tags to be installed, which will also satisfy `--prefer-lowest`.

PR #32, opened 2020-01-13
IDN domains were messed up with strtolower

mrothauer:
IDN domains like `körberl.taugl.online` were messed up to `k▒rberl.taugl.online`.

### DIFF
--- a/src/Uri.php
+++ b/src/Uri.php
@@ -21,6 +21,7 @@ use function is_numeric;
 use function is_object;
 use function is_string;
 use function ltrim;
+use function mb_strtolower;
 use function parse_url;
 use function preg_match;
 use function preg_replace;
@@ -462,7 +463,7 @@ class Uri implements UriInterface
 
         $this->scheme    = isset($parts['scheme']) ? $this->filterScheme($parts['scheme']) : '';
         $this->userInfo  = isset($parts['user']) ? $this->filterUserInfoPart($parts['user']) : '';
-        $this->host      = isset($parts['host']) ? strtolower($parts['host']) : '';
+        $this->host      = isset($parts['host']) ? mb_strtolower($parts['host']) : '';
         $this->port      = isset($parts['port']) ? $parts['port'] : null;
         $this->path      = isset($parts['path']) ? $this->filterPath($parts['path']) : '';
         $this->query     = isset($parts['query']) ? $this->filterQuery($parts['query']) : '';
--- a/src/Uri.php
+++ b/src/Uri.php
@@ -317,7 +317,7 @@ class Uri implements UriInterface
         }
 
         $new = clone $this;
-        $new->host = strtolower($host);
+        $new->host = mb_strtolower($host);
 
         return $new;
     }
--- a/src/Uri.php
+++ b/src/Uri.php
@@ -21,7 +21,6 @@ use function is_numeric;
 use function is_object;
 use function is_string;
 use function ltrim;
-use function mb_strtolower;
 use function parse_url;
 use function preg_match;
 use function preg_replace;
@@ -317,7 +316,7 @@ class Uri implements UriInterface
         }
 
         $new = clone $this;
-        $new->host = mb_strtolower($host);
+        $new->host = $this->lowercase($host);
 
         return $new;
     }
@@ -463,7 +462,7 @@ class Uri implements UriInterface
 
         $this->scheme    = isset($parts['scheme']) ? $this->filterScheme($parts['scheme']) : '';
         $this->userInfo  = isset($parts['user']) ? $this->filterUserInfoPart($parts['user']) : '';
-        $this->host      = isset($parts['host']) ? mb_strtolower($parts['host']) : '';
+        $this->host      = isset($parts['host']) ? $this->lowercase($parts['host']) : '';
         $this->port      = isset($parts['port']) ? $parts['port'] : null;
         $this->path      = isset($parts['path']) ? $this->filterPath($parts['path']) : '';
         $this->query     = isset($parts['query']) ? $this->filterQuery($parts['query']) : '';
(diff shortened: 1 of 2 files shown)


mrothauer:
After some debugging in my CakePHP 4.0.2 application I found out that strtolower messed up the `ö` in the uri. It would be great if you could confirm or deny that it is indeed a bug. Thank you.

snapshotpl:
@mrothauer rebase to master. After that tests should pass

mrothauer:
@snapshotpl does that mean, that there was a problem with UTF-8 encoding and it the fix was commited to the master branch in the meantime? Sorry for insisting a confirmation of the issue, if it's not a bug with laminas, I have to keep investigating elsewhere.

snapshotpl:
No. Your changes contains fix and unit test, however fails because there was a bug in master. Current master is fixed now, so you can check your changes fixes issue you post here.

mrothauer:
Ok now I got it... Tests pass!

mrothauer:
By adding the test `testWithHostWithGermanUmlaut` (without your suggested code-changing) it turned out, that the test passes. Should I change nevertheless change `strtolower` to `mb_strtolower` in the method `withHost`?

test/UriTest.php
```diff
@@ -136,6 +136,13 @@ public function testWithHostReturnsSameInstanceWithProvidedHostIsSameAsBefore()
         $this->assertSame('https://user:pass@local.example.com:3001/foo?bar=baz#quz', (string) $new);
     }
 
+    public function testWithHostWithGermanUmlaut()
+    {
+        $uri = new Uri('https://körberl.taugl.online');
+        $new = $uri->withHost('körberl.taugl.online');
+        $this->assertSame($uri, $new);
```
snapshotpl: Looks like wrong test because you put same value and `Url` is still same instance, so nothing changed. I suggest to change first host to something different and check result of `getHost()` like in `testUriWithGermanUmlaut`

mrothauer: Absolutely... I just fixed it and added mb_strtolower to method `withHost` - this time it's not working without it!

michalbundyra:
As we are using mb_* functions now we probably need to add `ext-mbstring` intro require section of composer.

@weierophinney What do you think? Is it ok to add it in bugfix version?

geerteltink:
> As we are using mb_* functions now we probably need to add ext-mbstring intro require section of composer.

You need to add it for sure. If the extension is not installed it throws an exception.

> Is it ok to add it in bugfix version?

I would say it should be a new major version, however it is a bugfix. At least make it a minor update because of the BC break.

michalbundyra:
@geerteltink 

> I would say it should be a new major version, however it is a bugfix. At least make it a minor update because of the BC break.

Adding requirements is not a BC Break.

geerteltink:
Why wouldn't it be a BC break? If someone hasn't installed the extension, he can't install the component. It's the same as increasing the PHP version. For zend framework we had the policy to do it only in a major update. However exceptions have been made / sneaked in. Not sure if the policy changed.

michalbundyra:
@geerteltink 
> Why wouldn't it be a BC break? If someone hasn't installed the extension, he can't install the component. 

Exactly, so this version will be not installed, because it's not compatible (because of requirements).

> It's the same as increasing the PHP version.

Not really. See the last paragraph there:
https://www.doctrine-project.org/2017/07/25/php-7.1-requirement-and-composer.html
This was quoted multiple times by @Ocramius and maybe he is even author of that ;)

> For zend framework we had...

Yes. But now we on Laminas and I believe it has been changed - so we can drop PHP Support in minor, as it is NOT a BC Break.

src/Uri.php
```diff
@@ -316,7 +317,7 @@ public function withHost($host) : UriInterface
         }
 
         $new = clone $this;
-        $new->host = strtolower($host);
+        $new->host = mb_strtolower($host);
```
weierophinney: Another option is to do feature detection, and use `mb_strtolower()` only if it's present on the system, and otherwise use `strtolower()`.

In most cases, if you are working with URIs that might contain multi-byte characters, the rest of your application likely does as well, so you'll have mbstring installed. If you don't, you'll hit the bug - and the recommendation will be "install mbstring".

Thoughts?

michalbundyra: @weierophinney I was just discussing that option with @geerteltink and he is worried about the performance.

Maybe we should add the check in bugfix version and in minor add the requirement?

weierophinney: We can cache the first call to test - just have an internal property, and a method that caches the lookup. These sorts of lookups are _very_ inexpensive though.

michalbundyra:
@mrothauer @snapshotpl 

I've started testing it locally in preparation for the release and ... it seems to be not needed.
See my last commit: d027283.
All tests pass with just `strtolower`.

Thoughts?

snapshotpl:
@michalbundyra TDD! that's why ever pull request should first contains falling test in first commit or/and  use mutation tests.

michalbundyra:
@mrothauer Closing as invalid.

mrothauer:
On my machine my test did not pass, that was enough proof for me.

> TDD! that's why ever pull request should first contains falling test in first commit or/and use mutation tests.

I'll keep that in mind in future, I'm sorry.

Xerkus:
I believe strtolower is affected by current locale.
I see no issue on my current `LC_CTYPE=en_US.UTF-8;`

@mrothauer can you provide output of `setlocale(LC_CTYPE, 0);` on the machine where output is getting garbled?

mrothauer:
> can you provide output of setlocale(LC_CTYPE, 0); on the machine where output is getting garbled?

`German_Germany.1252`

And if I set it to `de_DE.utf8` the test works! Great! Thank you!

Xerkus:
@weierophinney @michalbundyra do you think this environment issue is within the scope of the library? Reopen or wont-fix?

michalbundyra:
@Xerkus "wont-fix", imo.

Just checked guzzle/psr7 and they are doing just `strtolower` as well:
https://github.com/guzzle/psr7/blob/master/src/Uri.php#L591-L598

weierophinney:
> do you think this environment issue is within the scope of the library?

It may be worth a mention in the documentation.

PR #33, opened 2020-01-13
Use stable php-http/psr7-integration-tests

snapshotpl:
Fixes current master because `php-http/psr7-integration-tests` in master require phpunit in v8.

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes


michalbundyra:
Thanks, @snapshotpl!

PR #40, opened 2020-04-27
Bump PHP minimum to 7.3

weierophinney:
Per the policy agreed upon during the inaugural TSC meeting, this patch bumps the minimum supported PHP version to 7.3 for the next minor release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,26 +13,14 @@ env:
 matrix:
   fast_finish: true
   include:
-    - php: 7.1
+    - php: 7.3
       env:
         - DEPS=lowest
-    - php: 7.1
+    - php: 7.3
       env:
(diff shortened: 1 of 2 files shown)


michalbundyra:
@weierophinney It looks good, but... in case of security issues in the library we would need to provide fix to the previous current minor (previous once this is merged and released).
So until 30 Nov 2020 we have here two versions we need support:
- 2.2.x - security only
- 2.3.x - current/active support

Because of that, if there is no any feature we are using form PHP 7.3 I would prefer to keep it a bit slower, so we have minimum number of branches to support in case of security issues.

weierophinney:
Since I'm not making use of any 7.3-specific features, I'm retracting this. for now.

Issue #10, opened 2019-12-31
Invalid size provided for UploadedFile; must be an int.


weierophinney:
Hi guys, I'm working on an iOS project and I'm required to send documents, such as pdfs, to my laravel server. The issue is, every time I try to send a post request for a PDF, I get this error message: `Invalid size provided for UploadedFile; must be an int`. I'm at my wit's end trying to solve this issue. Any ideas?

---

Originally posted by @ImmanuelKannan at https://github.com/zendframework/zend-diactoros/issues/316

weierophinney:
That exception is thrown here https://github.com/zendframework/zend-diactoros/blob/167607290e79c396fd947cedeadccce0173c0e47/src/UploadedFile.php#L113-L115

That is all I can derive from the description, you need to be more specific. Stack traces, anything?

---

Originally posted by @Xerkus at https://github.com/zendframework/zend-diactoros/issues/316#issuecomment-405119085

weierophinney:
If laravel uses `ServerRequestFactory` to create request from php globals, try to do `var_dump($_FILES); exit;` with failing request and post result here

---

Originally posted by @Xerkus at https://github.com/zendframework/zend-diactoros/issues/316#issuecomment-405119510

weierophinney:
Increase upload_max_filesize in php.ini

;Maximum allowed size for uploaded files.
; http://php.net/upload-max-filesize
upload_max_filesize = 16M

Then
Restart Apache

---

Originally posted by @mrnaghibi at https://github.com/zendframework/zend-diactoros/issues/316#issuecomment-407001360

halfpastfouram:
While this error itself will no longer occur, almost the same problem still happens in 2.3.0. When receiving the uploaded files from a request:

```
Argument 2 passed to Laminas\Diactoros\UploadedFile::__construct() must be of the type integer, string given, called in src/functions/create_uploaded_file.php on line 38
```

This can be easily fixed in the code by casting both the file size and the error code as an integer. I'll be making an PR for this.

PR #44, opened 2020-07-03
Fixes laminas/laminas-diactoros#10

halfpastfouram:
Signed-off-by: halfpastfouram

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Really simple fix for https://github.com/laminas/laminas-diactoros/issues/10 with no BC breaks or anything. See https://github.com/laminas/laminas-diactoros/issues/10#issuecomment-653596086.

### DIFF
--- a/src/functions/create_uploaded_file.php
+++ b/src/functions/create_uploaded_file.php
@@ -32,8 +32,8 @@ function createUploadedFile(array $spec) : UploadedFile
 
     return new UploadedFile(
         $spec['tmp_name'],
-        $spec['size'],
-        $spec['error'],
+        intval($spec['size']),
+        intval($spec['error']),
         $spec['name'] ?? null,
         $spec['type'] ?? null
     );
--- a/src/functions/create_uploaded_file.php
+++ b/src/functions/create_uploaded_file.php
@@ -32,8 +32,8 @@ function createUploadedFile(array $spec) : UploadedFile
 
     return new UploadedFile(
         $spec['tmp_name'],
-        intval($spec['size']),
-        intval($spec['error']),
+        (int) $spec['size'],
+        $spec['error'],
         $spec['name'] ?? null,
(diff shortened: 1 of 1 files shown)


src/functions/create_uploaded_file.php
```diff
@@ -32,8 +32,8 @@ function createUploadedFile(array $spec) : UploadedFile
 
     return new UploadedFile(
         $spec['tmp_name'],
-        $spec['size'],
-        $spec['error'],
+        intval($spec['size']),
+        intval($spec['error']),
```
lcobucci, line 36: I think casting to would be a better alternative for this since there's no for the function call (we also don't need to perform any kind of base conversion).

```suggestion
        (int) $spec['size'],
        (int) $spec['error'],
```

I'm just afraid, though, that this hides the problem instead. The main question here is: how come these aren't integers?

halfpastfouram: Casting is fine. The underlying "problem" is that PHP returns the filesize as a string in the $_FILES global:

```
var_export($_FILES);

// array ( 'my_file' => array ( 'name' => 'myImage.png', 'type' => 'image/png', 'tmp_name' => '/tmp/phpr4VQcQ', 'error' => 0, 'size' => '50696', ), )
```

I now see that the error is actually an integer so only the size must be cast to integer. I'll change that.

Now, why the size is a string is not clear to me but it might have something to do with this bugfix from the past: https://bugs.php.net/bug.php?id=45124.

A 2GB file would be 2,147,483,648 bytes and max int (32 bit) would be 2,147,483,647. So it's possible they changed the size into a string because of that. This would be no problem with a 64 bit integer so I don't think it will be a problem just casting the string to an integer.

PR #45, opened 2020-08-03
Allow Streams to be instantiated using GD resources

settermjd:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

I've been working on a patch to laminas-barcode, which requires a Stream to be able to be created from a GD resource. However, the patch fails, as Diactoros only allows resource types to be of type stream. So, this patch expands the list of types allowed to instantiate a Stream to include GD types. I'm submitting this after discussing the idea with @weierophinney some time ago. I hope that I understood what he meant.


weierophinney:
Ping me for a review when it's ready, @settermjd !

Also, don't forget to use the `-s` switch to sign-off your commits! There's instructions in the contributors guide on how to use rebase to add them.

src/Stream.php
```diff
@@ -357,4 +362,15 @@ private function setStream($stream, string $mode = 'r') : void
...
+
+    /**
+     * Determine if a resource is one of the resource types allowed to instantiate a Stream
+     *
+     * @param resource $resource Stream resource.
+     * @return bool
+     */
+    protected function isAllowedStreamResourceType($resource)
+    {
+        return in_array(get_resource_type($resource), self::ALLOWED_TYPES);
```
samsonasik: may need strict check : 

```suggestion
        return in_array(get_resource_type($resource), self::ALLOWED_TYPES, true);
```

settermjd: Thanks for the feedback.

samsonasik: you seems push to wrong repo branch, should push to your fork branch instead of this repo branch

settermjd:
> Ping me for a review when it's ready, @settermjd !
> 
> Also, don't forget to use the `-s` switch to sign-off your commits! There's instructions in the contributors guide on how to use rebase to add them.

Ready for review. It's not a big patch, and I hope that I understood what you were saying in our previous discussion a little while ago.

src/Stream.php
samsonasik: please push to your fork ( as this PR ):

https://github.com/settermjd/laminas-diactoros/tree/expand-list-of-resource-types

instead of this repo branch:

https://github.com/laminas/laminas-diactoros/tree/expand-list-of-resource-types

settermjd: I must have done that by accident. Sorry.

settermjd: That said, I updated the PR to include strict type checking.

src/StreamTrait.php
```diff
@@ -0,0 +1,25 @@
...
+
+trait StreamTrait
+{
+    /**
+     * Determine if a resource is one of the resource types allowed to instantiate a Stream
+     *
+     * @param resource $resource Stream resource.
+     * @return bool
+     */
+    protected function isValidStreamResourceType($resource)
```
samsonasik: I think modifier can be `private` instead of `protected`

src/Stream.php
```diff
@@ -40,6 +40,13 @@
  */
 class Stream implements StreamInterface
 {
+    use StreamTrait;
+
+    /**
+     * A list of types that are allowed to instantiate a Stream
+     */
+    const ALLOWED_TYPES = ['stream', 'gd'];
```
samsonasik: modifier can be added, eg: private

settermjd: I don't have an objection, but would that really be necessary?

settermjd: Actually, it seems necessary to keep it public, as they're referenced in the call to `isValidStreamResourceType` in `StreamFactory`, on line 51.

samsonasik: /cc @weierophinney

settermjd: After a bit of thought, @samsonasik, I moved Stream::ALLOWED_TYPES to be a private variable in StreamTrait. Thanks for the suggestion and encouragement, and thanks to @weierophinney for further educating me.

src/StreamTrait.php
```diff
@@ -0,0 +1,30 @@
...
+     */
+    private $allowedTypes = ['gd', 'stream'];
+
+    /**
+     * Determine if a resource is one of the resource types allowed to instantiate a Stream
+     *
+     * @param resource $resource Stream resource.
+     * @return bool
+     */
+    private function isValidStreamResourceType($resource)
```
samsonasik: return declaration can be added `: bool` so no need `@return bool` docblock

src/Stream.php
```diff
@@ -40,6 +40,8 @@
  */
 class Stream implements StreamInterface
 {
+    use StreamTrait;
```
lcobucci: I'd suggest avoiding traits here.

I'd argue that only the stream should be responsible for the validation of its input.
I understand that the purpose of the patch is to extend the behaviour but I believe we could rethink this design.

src/StreamTrait.php
```diff
@@ -0,0 +1,30 @@
...
+declare(strict_types=1);
+
+namespace Laminas\Diactoros;
+
+trait StreamTrait
+{
+    /**
+     * A list of types that are allowed to instantiate a Stream
+     */
+    private $allowedTypes = ['gd', 'stream'];
```
lcobucci: This should rather be a private constant (which works great with the centralisation of the validation on the stream class).

Let's also remember that properties on traits are replicated in every instance, which might be a not so great usage of memory.

lcobucci (changes requested):

src/StreamTrait.php
settermjd: Thanks for spotting that.

src/Stream.php
settermjd: I've been pondering on this for a bit. Seems to make sense to me. Will refactor.

src/StreamTrait.php
settermjd: @lcobucci, I took your advice and rolled the functionality from StreamTrait into Stream. Thanks for your guidance.

src/StreamFactory.php
```diff
@@ -21,6 +21,7 @@
 
 class StreamFactory implements StreamFactoryInterface
 {
+
```
lcobucci: I know this is just coding standard, but since it hasn't been reported by CI... based on the rest of the codebase it looks like this shouldn't be added.

lcobucci (approved):
Just a teeny tiny thing! Thanks for your work here... things look much better and the stream class is much more coherent to me :+1:

lcobucci:
The failing integration test seems to come from `php-http/psr7-integration-tests`... it's doing external calls without retrying on failure, which makes the suite a bit flaky.

src/StreamFactory.php
settermjd: I'll check this out. Thanks for letting me know.

settermjd: Fixed!

lcobucci (approved):

weierophinney (approved):
:ship:

settermjd:
Big thanks to everyone for all your support on this PR.

PR #46, opened 2020-08-14
Allow installation on PHP 8.

### LINKED ISSUE #54: PHP 8.0 support
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes

#### Summary

To be prepared for the december release of PHP 8.0, this repository has some additional TODOs to be tested against the new major version.

In order to make this repository compatible, one has to follow these steps:

- [ ] Modify `composer.json` to provide support for PHP 8.0 by adding the constraint `~8.0.0`
- [ ] Modify `composer.json` to drop support for PHP less than 7.3
- [ ] Modify `composer.json` to implement phpunit 9.3 which supports PHP 7.3+
- [ ] Modify `.travis.yml` to ignore platform requirements when installing composer dependencies (simply add `--ignore-platform-reqs` to `COMPOSER_ARGS` env variable)
- [ ] Modify `.travis.yml` to add PHP 8.0 to the matrix (*NOTE*: Do not allow failures as PHP 8.0 has a feature freeze since 2020-08-04!)
- [ ] Modify source code in case there are incompatibilities with PHP 8.0

ADmad:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | yes
| QA            | no

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,38 +13,28 @@ env:
 matrix:
   fast_finish: true
   include:
-    - php: 7.1
+    - php: 7.3
       env:
         - DEPS=lowest
-    - php: 7.1
+    - php: 7.3
       env:
         - DEPS=latest
         - CS_CHECK=true
         - TEST_COVERAGE=true
-    - php: 7.2
-      env:
-        - DEPS=lowest
-    - php: 7.2
-      env:
-        - DEPS=latest
-    - php: 7.3
-      env:
-        - DEPS=lowest
-    - php: 7.3
-      env:
-        - DEPS=latest
     - php: 7.4
       env:
         - DEPS=lowest
     - php: 7.4
+      env:
+        - DEPS=latest
+    - php: 'nightly'
       env:
         - DEPS=latest
 
 before_install:
-  - if [[ $TEST_COVERAGE != 'true' ]]; then phpenv config-rm xdebug.ini || return 0 ; fi
+  - if [[ $TEST_COVERAGE != 'true' && $TRAVIS_PHP_VERSION != 'nightly' ]]; then phpenv config-rm xdebug.ini || return 0 ; fi
 
 install:
-  - travis_retry composer install $COMPOSER_ARGS
   - if [[ $DEPS == 'latest' ]]; then travis_retry composer update $COMPOSER_ARGS ; fi
   - if [[ $DEPS == 'lowest' ]]; then travis_retry composer update --prefer-lowest --prefer-stable $COMPOSER_ARGS ; fi
   - if [[ $TEST_COVERAGE == 'true' ]]; then travis_retry composer require --dev $COMPOSER_ARGS $COVERAGE_DEPS ; fi
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         }
     },
     "require": {
-        "php": "^7.1",
+        "php": "^7.3 || ^8.0",
         "laminas/laminas-zendframework-bridge": "^1.0",
         "psr/http-factory": "^1.0",
         "psr/http-message": "^1.0"
@@ -38,10 +38,11 @@
         "ext-dom": "*",
         "ext-gd": "*",
         "ext-libxml": "*",
-        "http-interop/http-factory-tests": "^0.5.0",
+        "http-interop/http-factory-tests": "^0.7.0",
         "laminas/laminas-coding-standard": "~1.0.0",
-        "php-http/psr7-integration-tests": "^1.0",
-        "phpunit/phpunit": "^7.5.18"
+        "php-http/psr7-integration-tests": "^1.1",
+        "phpspec/prophecy-phpunit": "^2.0",
+        "phpunit/phpunit": "^9.1"
     },
     "provide": {
         "psr/http-factory-implementation": "1.0",
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         }
     },
     "require": {
-        "php": "^7.3 || ^8.0",
+        "php": "^7.3 || ~8.0.0",
         "laminas/laminas-zendframework-bridge": "^1.0",
         "psr/http-factory": "^1.0",
         "psr/http-message": "^1.0"
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
         "ext-dom": "*",
         "ext-gd": "*",
         "ext-libxml": "*",
-        "http-interop/http-factory-tests": "^0.7.0",
+        "http-interop/http-factory-tests": "^0.8.0",
         "laminas/laminas-coding-standard": "~1.0.0",
         "php-http/psr7-integration-tests": "^1.1",
         "phpspec/prophecy-phpunit": "^2.0",
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -370,9 +370,14 @@ class Stream implements StreamInterface
      */
     private function isValidStreamResourceType($resource): bool
     {
-        return (
-            is_resource($resource) &&
-            in_array(get_resource_type($resource), self::ALLOWED_STREAM_RESOURCE_TYPES, true)
-        );
+        if (is_resource($resource)) {
+            return in_array(get_resource_type($resource), self::ALLOWED_STREAM_RESOURCE_TYPES, true);
+        }
+
+        if (PHP_VERSION_ID >= 80000 && $resource instanceof \GdImage) {
+            return true;
+        }
+
+        return false;
     }
 }
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -347,7 +347,7 @@ class Stream implements StreamInterface
         }
 
         if ($error) {
-            throw new Exception\InvalidArgumentException('Invalid stream reference provided');
+            throw new Exception\RuntimeException('Invalid stream reference provided');
         }
 
         if (! $this->isValidStreamResourceType($resource)) {
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 /phpunit.xml
 /test/TestAsset/.cache/
 /vendor/
+.phpunit.result.cache
--- a/src/Exception/RuntimeException.php
+++ b/src/Exception/RuntimeException.php
@@ -1,0 +1,15 @@
+<?php
+
+/**
+ * @see       https://github.com/laminas/laminas-diactoros for the canonical source repository
+ * @copyright https://github.com/laminas/laminas-diactoros/blob/master/COPYRIGHT.md
+ * @license   https://github.com/laminas/laminas-diactoros/blob/master/LICENSE.md New BSD License
+ */
+
+declare(strict_types=1);
+
+namespace Laminas\Diactoros\Exception;
+
+class RuntimeException extends \RuntimeException implements ExceptionInterface
+{
+}
--- a/test/RelativeStreamTest.php
+++ b/test/RelativeStreamTest.php
@@ -14,6 +14,7 @@ use Laminas\Diactoros\RelativeStream;
 use Laminas\Diactoros\Stream;
 use PHPUnit\Framework\TestCase;
 use Prophecy\Argument;
+use Prophecy\PhpUnit\ProphecyTrait;
 use RuntimeException;
 
 use const SEEK_SET;
@@ -23,6 +24,8 @@ use const SEEK_SET;
  */
 class RelativeStreamTest extends TestCase
 {
+    use ProphecyTrait;
+    
     public function testToString()
     {
         $decorated = $this->prophesize(Stream::class);
--- a/test/RelativeStreamTest.php
+++ b/test/RelativeStreamTest.php
@@ -25,7 +25,7 @@ use const SEEK_SET;
 class RelativeStreamTest extends TestCase
 {
     use ProphecyTrait;
-    
+
     public function testToString()
     {
         $decorated = $this->prophesize(Stream::class);
--- a/test/Request/SerializerTest.php
+++ b/test/Request/SerializerTest.php
@@ -54,8 +54,8 @@ class SerializerTest extends TestCase
             ->withBody($stream);
 
         $message = Serializer::toString($request);
-        $this->assertContains("POST /foo/bar HTTP/1.1\r\n", $message);
-        $this->assertContains("\r\n\r\n" . $body, $message);
+        $this->assertStringContainsString("POST /foo/bar HTTP/1.1\r\n", $message);
+        $this->assertStringContainsString("\r\n\r\n" . $body, $message);
     }
 
     public function testSerializesMultipleHeadersCorrectly()
@@ -67,8 +67,8 @@ class SerializerTest extends TestCase
             ->withAddedHeader('X-Foo-Bar', 'Bat');
 
         $message = Serializer::toString($request);
-        $this->assertContains("X-Foo-Bar: Baz", $message);
-        $this->assertContains("X-Foo-Bar: Bat", $message);
+        $this->assertStringContainsString("X-Foo-Bar: Baz", $message);
+        $this->assertStringContainsString("X-Foo-Bar: Bat", $message);
     }
 
     public function originForms()
--- a/test/RequestTest.php
+++ b/test/RequestTest.php
@@ -342,7 +342,7 @@ class RequestTest extends TestCase
         $request = new Request('http://example.com');
         $headers = $request->getHeaders();
         $this->assertArrayHasKey('Host', $headers);
-        $this->assertContains('example.com', $headers['Host']);
+        $this->assertStringContainsString('example.com', $headers['Host'][0]);
     }
 
     /**
@@ -421,7 +421,7 @@ class RequestTest extends TestCase
     {
         $request = new Request('http://example.com');
         $header = $request->getHeaderLine('host');
-        $this->assertContains('example.com', $header);
+        $this->assertStringContainsString('example.com', $header);
     }
 
     /**
--- a/test/Response/HtmlResponseTest.php
+++ b/test/Response/HtmlResponseTest.php
@@ -14,9 +14,12 @@ use InvalidArgumentException;
 use Laminas\Diactoros\Response\HtmlResponse;
 use PHPUnit\Framework\TestCase;
 use Psr\Http\Message\StreamInterface;
+use Prophecy\PhpUnit\ProphecyTrait;
 
 class HtmlResponseTest extends TestCase
 {
+    use ProphecyTrait;
+
     public function testConstructorAcceptsHtmlString()
     {
         $body = '<html>Uh oh not found</html>';
--- a/test/Response/JsonResponseTest.php
+++ b/test/Response/JsonResponseTest.php
@@ -128,7 +128,7 @@ class JsonResponseTest extends TestCase
         $contents = (string) $stream;
 
         $expected = json_encode($value, $defaultFlags);
-        $this->assertContains(
+        $this->assertStringContainsString(
             $expected,
             $contents,
             sprintf('Did not encode %s properly; expected (%s), received (%s)', $key, $expected, $contents)
--- a/test/Response/SerializerTest.php
+++ b/test/Response/SerializerTest.php
@@ -56,8 +56,8 @@ class SerializerTest extends TestCase
             ->withAddedHeader('X-Foo-Bar', 'Bat');
 
         $message = Serializer::toString($response);
-        $this->assertContains("X-Foo-Bar: Baz", $message);
-        $this->assertContains("X-Foo-Bar: Bat", $message);
+        $this->assertStringContainsString("X-Foo-Bar: Baz", $message);
+        $this->assertStringContainsString("X-Foo-Bar: Bat", $message);
     }
 
     public function testOmitsReasonPhraseFromStatusLineIfEmpty()
@@ -68,7 +68,7 @@ class SerializerTest extends TestCase
         $response->getBody()->write('Content!');
 
         $message = Serializer::toString($response);
-        $this->assertContains("HTTP/1.1 299\r\n", $message);
+        $this->assertStringContainsString("HTTP/1.1 299\r\n", $message);
     }
 
     public function testCanDeserializeBasicResponse()
--- a/test/Response/TextResponseTest.php
+++ b/test/Response/TextResponseTest.php
@@ -14,9 +14,12 @@ use InvalidArgumentException;
 use Laminas\Diactoros\Response\TextResponse;
 use PHPUnit\Framework\TestCase;
 use Psr\Http\Message\StreamInterface;
+use Prophecy\PhpUnit\ProphecyTrait;
 
 class TextResponseTest extends TestCase
 {
+    use ProphecyTrait;
+
     public function testConstructorAcceptsBodyAsString()
     {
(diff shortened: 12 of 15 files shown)


composer.json
```diff
@@ -44,7 +44,7 @@
         "http-interop/http-factory-tests": "^0.5.0",
         "laminas/laminas-coding-standard": "~1.0.0",
         "php-http/psr7-integration-tests": "^1.0",
-        "phpunit/phpunit": "^7.5.18"
+        "phpunit/phpunit": "^7.5.18 || ^9.3"
```
Ocramius: Should probably first upgrade the test suite

.travis.yml
```diff
@@ -39,9 +39,13 @@ matrix:
     - php: 7.4
       env:
         - DEPS=latest
+    - php: 'nightly'
+      env:
+        - DEPS=latest
+        - COMPOSER_ARGS="--no-interaction --ignore-platform-reqs"
```
Ocramius: If you added `^8.0`, then doing `--ignore-platform-reqs` here shouldn't be required

Ocramius (changes requested):

.travis.yml
ADmad: It's required temporarily as other dev deps like `php-http/psr7-integration-tests` and `laminas/laminas-zendframework-bridge` don't get installed without it as their deps also need to be updated to allow installing on PHP 8.

https://travis-ci.com/github/laminas/laminas-diactoros/jobs/372193687#L225

composer.json
ADmad: Currently diactoros is PHP 7.1+ while phpunit 9.3 is PHP 7.3+, so I was trying to see if both phpunit versions can be supported without bumping up diactoros' PHP version constraint :slightly_smiling_face:.

Ocramius: Yeah, we can bump to `^7.3` in the next minor

ADmad: Currently CakePHP 4 supports PHP 7.2+ and has diactoros as dependency. So unless really necessary we would prefer if diactoros does not goto 7.3+ just due to phpunit :)

Supporting phpunit 8.5 and 9.3 together with php 7.2+ isn't too difficult, we have done it https://github.com/cakephp/cakephp/pull/14885.

Ocramius: Cake can support an older version of diactoros, and then you'll be fine 👍

ADmad: Only if the older version supports PHP 8 too.

Ocramius: Why would it need to? If you pin to `^2.3.1`, you get `7.1.*` support, as well as PHP 8 support (which will happen with `2.4.0`, according to this patch)

ADmad:
Related PRs for dependencies:

https://github.com/laminas/laminas-zendframework-bridge/pull/69
https://github.com/php-http/psr7-integration-tests/pull/45
https://github.com/http-interop/http-factory-tests/pull/44

ADmad:
@Ocramius I have bumped PHP constraint to 7.3+ as per your feedback. Temporarily using my fork to run the test suite until `php-http/psr7-integration-tests` has a new release.

The two test failures aren't directly related to the PHP version bump and probably need to be addressed separately. 

The docblock of [StreamFactoryInterface::createStreamFromFile()](https://github.com/php-fig/http-factory/blob/1a2099a23a7bca67f13c0634723137a364a3ebf8/src/StreamFactoryInterface.php#L30) says `@throws \RuntimeException If the file cannot be opened.` and the test case correctly tests for that. But `Laminas\Diactoros\StreamFactory::createStreamFromFile()` always throws an instance of `Laminas\Diactoros\Exception\InvalidArgumentException`  (which extends `\InvalidArgumentException` if it fails to create a stream.

ADmad:
Regarding the failing tests this commit message has a quoted response for the PHP-FIG member: https://github.com/http-interop/http-factory-tests/pull/42/commits/c29f9c5a6b0c26c2aff25cbe46502aa9eeaf8157.

It's seems throwing a `RuntimeException` instead of `InvalidArgumentException` exception when failing to create a stream is recommend. So maybe the exception thrown by diactoros could be changed for next minor release?

Another option would be to see if http-interop is willing to just widen the asserted exception to `Exception` instead of `RuntimeException` as done for other tests in the commit linked above.

ADmad:
Can someone get the maintainers of `php-http/psr7-integration-tests` to merge or at least respond to https://github.com/php-http/psr7-integration-tests/pull/45 so that we can make progress on this PR?

weierophinney:
> Can someone get the maintainers of `php-http/psr7-integration-tests` to merge

I've bumped via a comment; crossing fingers for now.

weierophinney:
> Regarding the failing tests this commit message has a quoted response for the PHP-FIG member: [http-interop/http-factory-tests@c29f9c5](https://github.com/http-interop/http-factory-tests/commit/c29f9c5a6b0c26c2aff25cbe46502aa9eeaf8157).
> 
> It's seems throwing a `RuntimeException` instead of `InvalidArgumentException` exception when failing to create a stream is recommend. So maybe the exception thrown by diactoros could be changed for next minor release?
> 
> Another option would be to see if http-interop is willing to just widen the asserted exception to `Exception` instead of `RuntimeException` as done for other tests in the commit linked above.

Third option: we override that test case in our own test suite, to address the behavior.

ADmad:
> Third option: we override that test case in our own test suite, to address the behavior.

Not quite sure how to do that since `http-factory-tests`'s test suite is directly run https://github.com/laminas/laminas-diactoros/blob/e609ac206649cc51f00e9728314067848706dc33/phpunit.xml.dist#L7

ADmad:
BTW should this PR now target `2.5.x` branch? Given the min. PHP version bump I believe a new minor release would be warranted.

ADmad:
> BTW should this PR now target 2.5.x branch?

I took the liberty of doing so since the milestone for the PR was already changed to `2.5.0`.

ADmad:
I don't know your backward compatibility policy is but since these changes are going into a new minor release perhaps we can just update `Stream::setStream()` to throw `RuntimeException` instead of `InvalidArgument` exception to fix the testsuite failure.

ADmad:
I updated to use the new `php-http/psr7-integration-tests` release. Tweaked `Stream::isValidStreamResourceType()` to account for GD changes in PHP 8.

andypost:
Handy guide about changes https://php.watch/versions/8.0

ADmad:
>https://github.com/http-interop/http-factory-tests/commit/c29f9c5a6b0c26c2aff25cbe46502aa9eeaf8157
>
> It's seems throwing a RuntimeException instead of InvalidArgumentException exception when failing to create a stream is recommend. So maybe the exception thrown by diactoros could be changed for next minor release?

My recommendation would be to do the above to fix the test failures. I can make required changes if I get confirmation.

andypost:
Would be great to get in next minor release, upcoming Drupal 9.1 using it as dev-dependency, so would be great to declare full compatibility

froschdesign:
@andypost 
> Would be great to get in next minor release, upcoming Drupal 9.1 using it as dev-dependency, so would be great to declare full compatibility

Everyone is welcome to help!

andypost:
Checked 2 remaining failures - both caused by https://php.watch/versions/8.0/internal-function-exceptions

I see 2 ways to solve it

1. add condition to `\Laminas\Diactoros\Stream::setStream` before `fopen()` https://github.com/laminas/laminas-diactoros/blob/2.5.x/src/Stream.php#L345 - Drupal used to fix this way to keep backward compatibility https://www.drupal.org/project/drupal/issues/3177541

2. change expectations in https://github.com/http-interop/http-factory-tests/pull/45
- `\Interop\Http\Factory\StreamFactoryTestCase::testCreateStreamFromNonExistingFile()`
- `\Interop\Http\Factory\StreamFactoryTestCase::testCreateStreamFromInvalidFileName`

andypost:
Also filed issue for warnings https://github.com/php-http/psr7-integration-tests/pull/47

<details>
<summary>It allowed to pass 1 test</summary>

```
# composer test
> phpunit --colors=always
PHPUnit 9.4.2 by Sebastian Bergmann and contributors.

Warning:       Your XML configuration validates against a deprecated schema.
Suggestion:    Migrate your XML configuration using "--migrate-configuration"!

.............................................................   61 / 1101 (  5%)
.............................................................  122 / 1101 ( 11%)
.............................................................  183 / 1101 ( 16%)
.............................................................  244 / 1101 ( 22%)
.............................................................  305 / 1101 ( 27%)
.............................................................  366 / 1101 ( 33%)
.............................................................  427 / 1101 ( 38%)
.............................................................  488 / 1101 ( 44%)
.............................................................  549 / 1101 ( 49%)
.............................................................  610 / 1101 ( 55%)
.............................................................  671 / 1101 ( 60%)
.............................................................  732 / 1101 ( 66%)
.............................................................  793 / 1101 ( 72%)
.............................................................  854 / 1101 ( 77%)
.............................................................  915 / 1101 ( 83%)
.............................................................  976 / 1101 ( 88%)
............................................................. 1037 / 1101 ( 94%)
.....................................................F....... 1098 / 1101 ( 99%)
...                                                           1101 / 1101 (100%)

Time: 00:01.745, Memory: 12.00 MB

There was 1 failure:

1) Interop\Http\Factory\StreamFactoryTest::testCreateStreamFromNonExistingFile
Failed asserting that exception of type "Laminas\Diactoros\Exception\InvalidArgumentException" matches expected exception "PHPUnit\Framework\Error\Warning". Message was: "Invalid stream reference provided" at
/srv/src/Stream.php:350
/srv/src/Stream.php:65
/srv/src/StreamFactory.php:41
/srv/vendor/http-interop/http-factory-tests/test/StreamFactoryTestCase.php:95
.

FAILURES!
Tests: 1101, Assertions: 1834, Failures: 1.
Script phpunit --colors=always handling the test event returned with error code 1

```

</details>

andypost:
The last failed test is because test expects `RuntimeException` but `InvalidArgumentException` extends `LogicException` which is not runtime

andypost:
Following hack for  `http-interop/http-factory-tests` allowed to pass tests but looks instead of it the `Exception\InvalidArgumentException` should extend `\RuntimeException`

```diff
diff --git a/test/StreamFactoryTestCase.php b/test/StreamFactoryTestCase.php
index 4641788..5c9578a 100644
--- a/test/StreamFactoryTestCase.php
+++ b/test/StreamFactoryTestCase.php
@@ -86,7 +86,7 @@ abstract class StreamFactoryTestCase extends TestCase
         $filename = $this->createTemporaryFile();
         unlink($filename);
 
-        $this->expectException(RuntimeException::class);
+        $this->expectException(Exception::class);
         $stream = $this->factory->createStreamFromFile($filename);
     }
 
```

ADmad:
> Everyone is welcome to help!

@froschdesign I had been awaiting the Laminas team's response regarding how to deal with the `RuntimeException` vs `InvalidArgumentException` issue. With none forth coming I finally just went ahead and updated the code to throw `RuntimeException`.

andypost:
Both PRs for dependencies are accepted but no releases yet

composer.json
```diff
@@ -28,7 +28,7 @@
         }
     },
     "require": {
-        "php": "^7.1",
+        "php": "^7.3 || ^8.0",
```
boesing: Please use `~8.0.0` as we internally decided to go that way for PHP 8.0.

src/Stream.php
```diff
@@ -347,7 +347,7 @@ private function setStream($stream, string $mode = 'r') : void
         }
 
         if ($error) {
-            throw new Exception\InvalidArgumentException('Invalid stream reference provided');
+            throw new Exception\RuntimeException('Invalid stream reference provided');
```
boesing, line 350: As this is a BC break, I would prefer having some kind of "feature toggle" which allows upstream consumers to "enable" the `RuntimeException`

We could mark that feature toggle method as deprecated aswell as stating that with the next major, we will completely switch to `RuntimeException`.

Thoughts @weierophinney?

src/Stream.php
```diff
@@ -370,9 +370,14 @@ private function setStream($stream, string $mode = 'r') : void
...
     {
-        return (
-            is_resource($resource) &&
-            in_array(get_resource_type($resource), self::ALLOWED_STREAM_RESOURCE_TYPES, true)
-        );
+        if (is_resource($resource)) {
+            return in_array(get_resource_type($resource), self::ALLOWED_STREAM_RESOURCE_TYPES, true);
+        }
+
+        if (PHP_VERSION_ID >= 80000 && $resource instanceof \GdImage) {
```
boesing, line 377: As this is the only change in here, I have a few questions:

- does all of these functions work on that `GdImage` object?
  - `fclose`
  - `stream_get_contents`
  - `fread`
  - `stream_get_meta_data`

If the answer for any of these functions is "no", we need to find a completely new way to make the `Stream`-implementation compatible with the new `GdImage` object as we are using `fclose` in `Stream::close`, `stream_get_contents` in `Stream::getContents` and `stream_get_metadata` in some of the other methods...

boesing (changes requested):

composer.json
alexpott: @boesing is there a record of the discussion about using ~8.0.0? That means when 8.1 comes out any of your dependencies will require a new release of their laminas dependency for them to be installable even if there are no code changes required to the laminas library. This feels a step backwards. I think the approach recently taken by Symfony to change their constraints to `>=` is the way to go but I can understand libraries choosing to go with `^8.0` - but ~8.0.0 feels very restrictive.

Ocramius: >  That means when 8.1 comes out any of your dependencies will require a new release of their laminas dependency for them to be installable even if there are no code changes required to the laminas library.

This is intentional: PHP is too unstable to pin to a major release, and indeed, restricting is intentional. As for symfony, I can only say that jumping off the bridge because somebody told you to do it is not a wise approach to decision-making.

If all is good, next year, it is just a matter of bumping dependency ranges: until then, there is no guarantee that 8.1 will be compatible :-)

As for the meeting where this was decided, I think it was this one: https://github.com/laminas/technical-steering-committee/blob/main/meetings/minutes/2020-09-07-TSC-Minutes.md#hacktoberfest

Can't find the precise point in any meeting notes though - would need to check the `#tsc-meeting` channel on https://laminas.slack.com/

src/Stream.php
ADmad: I checked locally on PHP 7.3. The following test cases

```
    public function testIsReableWithGdResource()
    {
        $resource = imagecreate(1, 1);
        $stream   = new Stream($resource);
        $this->assertTrue($stream->isReadable());
    }

    public function testCloseWithGdResource()
    {
        $resource = imagecreate(1, 1);
        $stream   = new Stream($resource);
        $this->assertNull($stream->close());
    }

```
result in 
```
1) LaminasTest\Diactoros\StreamTest::testIsReableWithGdResource
stream_get_meta_data(): supplied resource is not a valid stream resource

2) LaminasTest\Diactoros\StreamTest::testCloseWithGdResource
fclose(): supplied resource is not a valid stream resource
```

So the existing code is already broken and doesn't work with image resources.

ADmad: So if you do want to support using image resources it should be sorted out separately as fixing existing broken functionality is outside the scope of this PR.

composer.json
alexpott: @Ocramius I'm not jumping off a bridge because somebody told me too. I'm being stopped sending out scouts across a bridge because someone is unnecessarily putting a stop sign in the way.

There are a number of reasons why I think constraining the maximum version of PHP a library can run on is a not healthy for the ecosystem. Here are a some:
* Increases friction to early adopters from testing your code and creating PRs to fix it
* Increases friction to dependencies testing against new PHP versions. They have to jump through hoops and if nothing is broken they then require a release from Laminas in order to run tests. It makes the ecosystem form a long chain and, if you are on the end of the chain, you have a lot waiting to do.
* Something hand wavy about software freedoms, gophp5, and Postel's law. It just feels wrong.
* Composer's PHP constraints don't really work this way anyway... if there's something like `"platform": { "php": 7.4.99 }` in the root composer.json then, even though I'm running PHP 8, composer will still get laminas dependencies and not inform me a project is incompatible.

I can understand limiting to major versions ie something like `^7.0` (although I disagree with it) but `~8.0.0` increases the friction for laminas's dependencies due to a vague "PHP is too unstable to pin to a major release". I've worked on making Drupal 7.0, 7.1, 7.2, 7.4 and 8.0 compatible. It's a very large codebase and has many dependencies. PHP is generally the least of the problems. The biggest problem in maintaining compatibility against multiple PHP versions at the moment is PHPUnit.

For example, any dependency of laminas-feed that wants to be compatible with PHP 8.0 is currently waiting on https://github.com/laminas/laminas-feed/pull/28 - which is only about code quality - and the PHP 8.0 compatibility PR only stopped PHP from issuing a deprecation message (a good thing to be sure) BUT now in order to be PHP 8 compatible your dependencies have to upgrade from 2.12.x to 2.13.0 - when there is no new functionality. There's a similar story for laminas-escaper and laminas-diactoros.

ADmad: > The biggest problem in maintaining compatibility against multiple PHP versions at the moment is PHPUnit.

This! For e.g. the only reason PHP version contraint for this package needed to be updated to `^7.3` is because of PHPUnit.

I don't know in what utopia Sebastian lives where the whole PHP ecosystem gets updated as soon as active support for a PHP version ends.

src/Stream.php
boesing: You are right, I just wanted to verify that we have this covered.
I will try to keep in mind that I create an Issue regarding this.

boesing (approved):

composer.json
boesing: No one stops you from using composer v2 with `--ignore-platform-req=php`.
So what is the real pain you want to express?

We just want to specify every PHP version as tested. This can only work if we have active tests against that version. So every user of laminas can be 100% sure (as of PHP 8.0) that the code is tested under the current version the user executes `composer update/install`.

Imho, thats a huge promise to all laminas users and gives a maximum of "safety". Bugs can rise even with tested versions but at least it is compatible and no-one has to check if a vendor package of laminas/mezzio/laminas-api-tool is compatible with PHP 8.
Users are not beta-testers for new PHP versions but actually thats what symfony and most other libs are doing. They let their users beta-test a probably unstable version of their package. Imho, thats the point what we should talk about and not about "friction".

**And this is what I call a healthy PHP ecosystem**

froschdesign: @alexpott
> For example, any dependency of laminas-feed that wants to be compatible with PHP 8.0 is currently waiting on laminas/laminas-feed#28 - which is only about code quality

Unfortunately this is not entirely correct. This pull request is completed but [the commits needs a correct signed off](https://github.com/laminas/laminas-feed/pull/28/checks?check_run_id=1336868041) by the creator of the pull request.
But I hope that we can solve the problem on laminas-feed quickly.

alexpott: I've just found https://github.com/php-http/psr7-integration-tests/pull/45 and I'm more than a bit bemused. But to find the laminas project asking for PHP 8 compatibility on project because it is holding back the ecosystem is a bit much. The points made are exactly the points I'm making about needing compatibility before a release in order to be able to test before. I quote...
> bump
>
> RC2 just dropped, so PHP 8 is reaching relative stability. A number of PSR-7 providers are trying to ensure they are ready (e.g., Diactoros, Slim), but we are all currently blocked until this is merged.

Diactoros has been sitting on approved merge requests for days, and similarly for Feeds (who's PHP 8 compatibility release is blocked on a new integration with Psalm...). This means that unless Drupal makes exceptions to its release management policies it'll be June 2021 before we have an officially PHP 8 compatible release.

Also given the time its taken to get 8.0.0 compatibility done setting your constraint to ~8.0.0 does not fill your dependencies with confidence that you'll 8.1.0 compatible releases in place in time for your dependencies to be 8.1.0 compatible.

dereuromark:
Projects and libraries need this PR to get merged in order to avoid any --ignore-platform-req hacks when finalizing PHP 8 compatibility.
Is there anything else needed to get this done at this point?

src/Stream.php
weierophinney: I'd consider this a bugfix, but call it out in the "Changed" section of the changelog, and note the _why_ of the change ("to comply with PSR-7 integration test suites")

composer.json
weierophinney: These are all valid view points. However, there's an aspect that those _consuming_ the library are missing from those _maintaining_ the library: Using `^` constraints makes maintaining the library far harder.

How?

Because every time we get a new minor or major PHP version, we have to increase our test matrix. And every time we increase the test matrix, we generally have to revisit the PHPUnit versions we use and the syntax each uses to ensure that tests will actually run. This makes even simple contributions often more complex because new failures arise that were not present before; additionally, if the contributor does not have every version of PHP currently supported by the library on hand, they may not even know there are test failures until CI/CD runs... which, when the matrix gets large, take longer and longer to complete. And we all know that when we have to wait for results, we tend to get bored and move on.

Yes, in an **ideal** world, there would be no BC breaks or deprecations within a given minor release series, whether that's PHP or libraries you use. But in point of fact, these arise **all the time**. That's the origin of the "PHP is too unstable to pin to a major release'" statement. Every single minor release that comes out, we MUST do a cycle to test our libraries against it, because inevitably some new deprecation or change **does** lead to breakage, whether that's simply a test failing because of a new deprecation notice, or, worse, **users** identifying issues. When a user jumps to a new version and suddenly the library does not work, that immediately becomes **our** problem, but now there's urgency to it.

On top of that, there's the question of how long we support a given minor or major release of a library. If the library has versions 1.0, 1.1, 1.2, 2.0, 2.1, 3.0, and 3.1, and all of them have a minimum supported version of `^7.0`, does that mean we have to provide bugfixes and security fixes to _all_ of these branches? or just the latest minor releases of each major branch (e.g., 1.2, 2.1, and 3.1)? or current major and previous major?

By narrowing the window of PHP versions we support (via the `~` constraint, or bumping the minimum minor we support), we make the maintenance story far easier:

- We provide security patches to branches where the minimum supported PHP version is still supported by community PHP.
- We can have a smaller number of PHPUnit versions we have to be compatible with.
- There's a smaller test matrix we need to fulfill.
- When a new minor or major version of PHP is released, we create a new minor release that changes the PHP constraints, to signal that we have actually _tested_ it on that version, versus _assumed_ it would remain compatible.

weierophinney:
> Projects and libraries need this PR to get merged in order to avoid any --ignore-platform-req hacks when finalizing PHP 8 compatibility.
> Is there anything else needed to get this done at this point?

Reviewing today, and will likely get a release out.

weierophinney (approved):
:ship:

composer.json
alexpott: > we generally have to revisit the PHPUnit versions we use and the syntax each uses to ensure that tests will actually run.

Fortunately the world has changed... https://twitter.com/s_bergmann/status/1328197689029386240

ADmad: Finally, Hallelujah!

GrahamCampbell:
The upstream change is wrong. You need to catch the ValueError from fopen on PHP 8.

froschdesign:
@GrahamCampbell 
Please open a new issue report because this pull request is already closed. Thanks in advance! 👍

GrahamCampbell:
A fix will be made to the upstream package, which will break this package's tests when it is tagged.

PR #59, opened 2020-12-10
psalm integration

### LINKED ISSUE #53: Psalm integration
### Feature Request

|    Q        |   A
|------------ | ------
| QA    | yes

#### Summary

As decided during the [Technical-Steering-Committee Meeting on August 3rd, 2020](https://github.com/laminas/technical-steering-committee/blob/main/meetings/minutes/2020-08-03-TSC-Minutes.md#vote-on-static-analysis-tooling), Laminas wants to implement [vimeo/psalm](https://github.com/vimeo/psalm) in all packages.

Implementing psalm is quite easy.

#### Required
- [ ] Create a `.psalm.xml.dist` in the project root
- [ ] Copy and paste the contents from this [psalm.xml.dist](https://github.com/mezzio/repo-template/blob/6124090610ef1c6d6b0c80a063cafacc3ec68f38/psalm.xml.dist)
- [ ] Run `$ composer require vimeo/psalm`
- [ ] Run `$ vendor/bin/psalm --set-baseline=psalm-baseline.xml`
- [ ] Add a composer script `static-analysis` with the command `psalm --shepherd --stats`
- [ ] Add a new line to `script:` in `.travis.yml`: `- if [[ $TEST_COVERAGE == 'true' ]]; then composer static-analysis ; fi`
- [ ] Remove phpstan from the project (`phpstan.neon.dist`, `.travis.yml` entry, `composer.json` `require-dev` and `scripts`)

##### Optional
- [ ] Fix as many psalm errors as possible.

snapshotpl:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Closes #53

### DIFF
--- a/psalm.xml.dist
+++ b/psalm.xml.dist
@@ -1,0 +1,33 @@
+<?xml version="1.0"?>
+<psalm
+    totallyTyped="true"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xmlns="https://getpsalm.org/schema/config"
+    xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
+    errorBaseline="psalm-baseline.xml"
+>
+    <projectFiles>
+        <directory name="src"/>
+        <directory name="test"/>
+        <ignoreFiles>
+            <directory name="vendor"/>
+        </ignoreFiles>
+    </projectFiles>
+
+    <issueHandlers>
+        <InternalMethod>
+            <errorLevel type="suppress">
+                <referencedMethod name="PHPUnit\Framework\MockObject\Builder\InvocationMocker::method"/>
+            </errorLevel>
+            <errorLevel type="suppress">
+                <referencedMethod name="PHPUnit\Framework\MockObject\Builder\InvocationMocker::willReturn"/>
+            </errorLevel>
+            <errorLevel type="suppress">
+                <referencedMethod name="PHPUnit\Framework\MockObject\Builder\InvocationMocker::with"/>
+            </errorLevel>
+        </InternalMethod>
+    </issueHandlers>
+    <plugins>
+        <pluginClass class="Psalm\PhpUnitPlugin\Plugin"/>
+    </plugins>
+</psalm>
--- a/psalm.xml.dist
+++ b/psalm.xml.dist
@@ -18,11 +18,7 @@
         <InternalMethod>
             <errorLevel type="suppress">
                 <referencedMethod name="PHPUnit\Framework\MockObject\Builder\InvocationMocker::method"/>
-            </errorLevel>
-            <errorLevel type="suppress">
                 <referencedMethod name="PHPUnit\Framework\MockObject\Builder\InvocationMocker::willReturn"/>
-            </errorLevel>
-            <errorLevel type="suppress">
                 <referencedMethod name="PHPUnit\Framework\MockObject\Builder\InvocationMocker::with"/>
             </errorLevel>
         </InternalMethod>
--- a/src/AbstractSerializer.php
+++ b/src/AbstractSerializer.php
@@ -118,6 +118,8 @@ abstract class AbstractSerializer
 
     /**
      * Serialize headers to string values.
+     *
+     * @psalm-param array<string, string[]> $headers
      */
     protected static function serializeHeaders(array $headers) : string
     {
@@ -135,7 +137,7 @@ abstract class AbstractSerializer
     /**
      * Filter a header name to wordcase
      */
-    protected static function filterHeader($header) : string
+    protected static function filterHeader(string $header) : string
     {
         $filtered = str_replace('-', ' ', $header);
         $filtered = ucwords($filtered);
--- a/src/AbstractSerializer.php
+++ b/src/AbstractSerializer.php
@@ -136,8 +136,10 @@ abstract class AbstractSerializer
 
     /**
      * Filter a header name to wordcase
+     *
+     * @param string $header
      */
-    protected static function filterHeader(string $header) : string
+    protected static function filterHeader($header) : string
     {
         $filtered = str_replace('-', ' ', $header);
         $filtered = ucwords($filtered);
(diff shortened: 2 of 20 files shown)


src/MessageTrait.php
```diff
@@ -37,13 +37,17 @@ trait MessageTrait
      * List of all registered headers, as key => array of values.
      *
      * @var array
+     *
+     * @psalm-var array<array-key, array<array-key, string>>
```
boesing:
```suggestion
     * @psalm-var array<non-empty-string, list<string>>
```

src/MessageTrait.php
```diff
@@ -37,13 +37,17 @@ trait MessageTrait
...
      */
     protected $headers = [];
 
     /**
      * Map of normalized header name to original name used to register header.
      *
      * @var array
+     *
+     * @psalm-var array<string, string>
```
boesing:
```suggestion
     * @psalm-var array<non-empty-string, non-empty-string>
```

src/MessageTrait.php
```diff
@@ -110,6 +114,8 @@ public function withProtocolVersion($version) : MessageInterface
      *
      * @return array Returns an associative array of the message's headers. Each
      *     key MUST be a header name, and each value MUST be an array of strings.
+     *
+     * @psalm-return array<array-key, array<array-key, string>>
```
boesing:
```suggestion
     * @psalm-return array<non-empty-string, list<string>>
```

src/Response.php
```diff
@@ -37,6 +37,8 @@ class Response implements ResponseInterface
      * Map of standard HTTP status code/reason phrases
      *
      * @var array
+     *
+     * @psalm-var array<int, string>
```
boesing:
```suggestion
     * @psalm-var array<positive-int, non-empty-string>
```

boesing (changes requested):
I prefer having `@var array<x,y>` annotations over `@psalm-var` as these are valid in most IDEs as of now.
Afaik, `list<x>` is still not supported by most of the IDEs.

I know that @Ocramius has some opinions on that.

In my company, we are adding:

```php
/**
 * @var array<int,string>
 * @psalm-var list<string>
 */
```

Ocramius:
> I prefer having `@var array<x,y>` annotations over `@psalm-var` as these are valid in most IDEs as of now.
Afaik, `list<x>` is still not supported by most of the IDEs.

If you want IDE support, you can add `@var` and `@psalm-var` in the same block.

Careful with removing `list<T>`! `list<T>` is a much stricter (better) type than `array<int, T>`, so if you really want IDE support, do not remove it, but add a `@var` block

psalm.xml.dist
```diff
@@ -0,0 +1,33 @@
...
+        <ignoreFiles>
+            <directory name="vendor"/>
+        </ignoreFiles>
+    </projectFiles>
+
+    <issueHandlers>
+        <InternalMethod>
+            <errorLevel type="suppress">
+                <referencedMethod name="PHPUnit\Framework\MockObject\Builder\InvocationMocker::method"/>
+            </errorLevel>
```
Ocramius: I don't remember if this is copied from somewhere, but these `<errorLevel` blocks can be merged together 👍

Ocramius: Also: not a blocker 👍

Ocramius (approved):
I think @boesing's suggestions cover most of what needs to be improved at type level: once that's through, 🚢

weierophinney:
@snapshotpl I've rebased against current 2.6.x, which includes the cutover to GHA. The Psalm check is failing, and it looks like there's still some feedback that needs to be addressed; ping me when done, and I'll do a final review.

lcobucci:
> Afaik, list<x> is still not supported by most of the IDEs.

Everything works on PHPStorm 2021.1 here (both on `@var` and `@param`) :+1:

snapshotpl:
@weierophinney done

src/AbstractSerializer.php
```diff
@@ -141,7 +143,7 @@ protected static function serializeHeaders(array $headers) : string
     /**
      * Filter a header name to wordcase
      */
-    protected static function filterHeader($header) : string
+    protected static function filterHeader(string $header) : string
```
boesing: Adding native type-hint to non-private methods on a non-final class is a BC break due to mismatching signature.

src/CallbackStream.php
```diff
@@ -104,7 +104,7 @@ public function isSeekable() : bool
     /**
      * {@inheritdoc}
      */
-    public function seek($offset, $whence = SEEK_SET)
+    public function seek($offset, $whence = SEEK_SET) : void
```
boesing: class is not final, thus this is a BC break

src/MessageTrait.php
```diff
@@ -138,20 +144,20 @@ public function hasHeader($header) : bool
      * If the header does not appear in the message, this method MUST return an
      * empty array.
      *
-     * @param string $header Case-insensitive header field name.
+     * @param string $name Case-insensitive header field name.
```
boesing: Is there a specific reason why you would rename this? Just curious tho.
I am pretty sure that this might be a BC break aswell, as we already have PHP 8 support which added method calls with named arguments which will break if you change this.

But as we are implementing psalm, I'd prefer having the `@no-named-arguments` in here. Maybe we move this change to the next major even tho I don't understand why this change is here in the first place.

src/Request.php
```diff
@@ -57,10 +57,10 @@ public function getHeaders() : array
     /**
      * {@inheritdoc}
      */
-    public function getHeader($header) : array
+    public function getHeader($name) : array
```
boesing: Is there a specific reason why you would rename this? Just curious tho.
I am pretty sure that this might be a BC break aswell, as we already have PHP 8 support which added method calls with named arguments which will break if you change this.

But as we are implementing psalm, I'd prefer having the `@no-named-arguments` in here. Maybe we move this change to the next major even tho I don't understand why this change is here in the first place.

boesing (dismissed):

weierophinney (approved):

PR #60, opened 2021-03-16
2.4.x - Allow installation on PHP 8 

jrbasso:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | yes
| QA            | no

### Description

Same as #46 for 2.4.x series. This will allow the library to be installed in wider ranger of PHP versions since 2.5 can't be installed in versions below 7.3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,30 +13,27 @@ env:
 matrix:
   fast_finish: true
   include:
-    - php: 7.1
+    - php: 7.2
       env:
         - DEPS=lowest
-    - php: 7.1
+    - php: 7.2
+      env:
+        - DEPS=latest
+    - php: 7.3
+      env:
+        - DEPS=lowest
+    - php: 7.3
       env:
         - DEPS=latest
         - CS_CHECK=true
         - TEST_COVERAGE=true
-    - php: 7.2
-      env:
-        - DEPS=lowest
-    - php: 7.2
-      env:
-        - DEPS=latest
-    - php: 7.3
-      env:
-        - DEPS=lowest
-    - php: 7.3
-      env:
-        - DEPS=latest
     - php: 7.4
       env:
         - DEPS=lowest
     - php: 7.4
+      env:
+        - DEPS=latest
+    - php: 8.0
       env:
         - DEPS=latest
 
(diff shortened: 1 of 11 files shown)


.travis.yml
```diff
@@ -13,14 +13,6 @@ env:
...
   fast_finish: true
   include:
-    - php: 7.1
-      env:
-        - DEPS=lowest
-    - php: 7.1
-      env:
-        - DEPS=latest
-        - CS_CHECK=true
-        - TEST_COVERAGE=true
```
weierophinney, line 23: We cannot drop versions from testing that are explicitly supported via the package constraints. You either need to bump the PHP constraint to `^7.2`, or keep the above.

.travis.yml
```diff
@@ -33,18 +25,22 @@ matrix:
...
         - DEPS=latest
+        - CS_CHECK=true
+        - TEST_COVERAGE=true
     - php: 7.4
       env:
         - DEPS=lowest
     - php: 7.4
       env:
         - DEPS=latest
+    - php: nightly
```
weierophinney: `8.0` is the correct choice here, as Travis's "nightly" target is now 8.1 development.

.travis.yml
```diff
@@ -33,18 +25,22 @@ matrix:
...
+      env:
+        - DEPS=latest
 
 before_install:
-  - if [[ $TEST_COVERAGE != 'true' ]]; then phpenv config-rm xdebug.ini || return 0 ; fi
+  - if [[ $TEST_COVERAGE != 'true' && $TRAVIS_PHP_VERSION != 'nightly' ]]; then phpenv config-rm xdebug.ini || return 0 ; fi
```
weierophinney: This is redundant, as `TEST_COVERAGE` is not set for that element in the matrix.

weierophinney (commented):
I've noted some problems with the Travis configuration.

However, I think the more important concern I have is: why would we expand the PHP versions supported on this particular release series?

The Laminas security window policy indicates that we only support (a) the current release branch, and (b) any release branch where the minimum supported PHP version is still supported by php.net.

The 2.4 series falls outside that realm. On top of that, if a user is on 7.3 or 7.4 currently, _they can already upgrade to the 2.5 series_; if they are on PHP 8, installation will give them 2.5+. Expanding what we support on the 2.4 series is unnecessary, as users of later versions of PHP can already get a later version.

.travis.yml
jrbasso: PHPUnit 8.x doesn't support PHP 7.1 and PHPUnit 7.x doesn't support 8.x. What do you recommend in this case?

.travis.yml
jrbasso: OK. I will update.

.travis.yml
jrbasso: Makes sense. Will change.

jrbasso:
The main reason for supporting PHP 8 in this old minor version is because our system uses PHP 7.2 and we are migrating to 8.0. Since our app uses `composer.lock` we can't change between the 2 different environments. Supporting PHP 8.0 on 2.4 will allow us to use the same package independent of the PHP version. Once we fully migrate to PHP 8 we can upgrade to the version 2.5+.

Otherwise we would need to migrate all our systems to something like PHP 7.4 first and then 8.0, which almost doubles the work required for the upgrade.

jrbasso:
To avoid changing the library and the minimum PHP tested version because of PHPUnit I will be closing it. Since the code itself is compatible with PHP 8 for the most part, we can specify the platform version in the composer while we are doing the transition.

PR #61, opened 2021-04-09
Add GHA CI workflow

internalsystemerror:
Converts CI workflow from Travis to GHA using: https://gist.github.com/weierophinney/b003e50c3c2667d08076caf31ebd36a4


.laminas-ci.json
```diff
@@ -0,0 +1,8 @@
+{
+    "extensions": [
+        "curl",
+        "dom",
+        "gd",
+        "libxml"
```
boesing: Afaik, `libxml` is bundled with PHP.
Same for `gd`, `dom` and `curl`.

Try to drop these to see if build passes. 👍🏼

README.md
```diff
@@ -1,7 +1,6 @@
 # laminas-diactoros
 
-[![Build Status](https://travis-ci.com/laminas/laminas-diactoros.svg)](https://travis-ci.com/laminas/laminas-diactoros)
-[![Coverage Status](https://coveralls.io/repos/github/laminas/laminas-diactoros/badge.svg)](https://coveralls.io/github/laminas/laminas-diactoros)
+[![Build Status](https://github.com/laminas/laminas-diactoros/workflows/Continuous%20Integration/badge.svg)](https://github.com/laminas/laminas-diactoros/actions?query=workflow%3A"Continuous+Integration")
```
boesing: Could you apply the changes as described here? I think thats more future proof than using the workflow name.
https://docs.github.com/en/actions/managing-workflow-runs/adding-a-workflow-status-badge#using-the-workflow-file-name

boesing (dismissed):

README.md
internalsystemerror: done

.laminas-ci.json
internalsystemerror: You are correct in that curl, dom and libxml come bundled but GD is not. I have updated the configuration to reflect this.

PR #66, opened 2021-05-17
Allow strict checks for Content-* headers

### LINKED ISSUE #63: Environment variables prefixed with `CONTENT_` may overwrite HTTP-Headers provided by the client
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | all

#### Summary

Environment variables prefixed with `CONTENT_` may overwrite HTTP-Headers provided by the client

#### Current behavior

This may be entirely expected behavior  but I have not found a spec or warning in the documentation.
It's a least surprising to find unrelated environment variables as request header and potentially a security problem when comparing the env-variable against the header during authentication.

##### Environment variables prefixed with `CONTENT_` will be populated as http-headers:

I stumbled about this behavior in a project with the following  `.env`-file:

```
CONTENT_API_PASSWORD=password-for-content-api
```

The `.env` file is loaded into the $_SERVER superglobal via [`Dotenv\Dotenv::createImmutable()`](https://github.com/vlucas/phpdotenv/blob/v5.3.0/src/Dotenv.php#L148).

The request object is created via `\Laminas\Diactoros\ServerRequestFactory::fromGlobals`.
The request object now contains `content-api-password`  as request-header ` $request->getHeader('content-api-password')`.

##### Environment variables prefixed with `CONTENT_` max overwrite client provided HTTP-Headers:

If the same header name is provided by an http-client the `$_SERVER` superglobal contains a `HTTP_CONTENT_API_PASSWORD` key. 

In the current implementation the order is relevant. Whatever key is defined later in `$_SERVER` will be used to populate the HTTP-Header.

#### How to reproduce

```php
        $server = [
            'HTTP_CONTENT_API_PASSWORD' => 'password_from_header',
            'CONTENT_API_PASSWORD' => 'password_from_env',
        ];

        $request = ServerRequestFactory::fromGlobals($server);

        $this->assertSame('password_from_env', $request->getHeader('content-api-password')[0]);


        $server = [
            'CONTENT_API_PASSWORD' => 'password_from_env',
            'HTTP_CONTENT_API_PASSWORD' => 'password_from_header',
        ];

        $request = ServerRequestFactory::fromGlobals($server);

        $this->assertSame('password_from_header', $request->getHeader('content-api-password')[0]);
```

#### Expected behavior

```php
        $server = [
            'HTTP_CONTENT_API_PASSWORD' => 'password_from_header',
            'CONTENT_API_PASSWORD' => 'password_from_env',
        ];

        $request = ServerRequestFactory::fromGlobals($server);

        $this->assertSame('password_from_header', $request->getHeader('content-api-password')[0]);


        $server = [
            'CONTENT_API_PASSWORD' => 'password_from_env',
            'HTTP_CONTENT_API_PASSWORD' => 'password_from_header',
        ];

        $request = ServerRequestFactory::fromGlobals($server);

        $this->assertSame('password_from_header', $request->getHeader('content-api-password')[0]);
```

weierophinney:
Previously, we would check any key starting with `CONTENT_` in the `$_SERVER` array, and treat it as a header.  However, per #63, this is incorrect, and we should only look for `CONTENT_TYPE`, `CONTENT_LENGTH`, and `CONTENT_MD5`.  To keep backwards compatibility, this patch implements a switch, using the ENV variable LAMINAS_DIACTOROS_STRICT_CONTENT_HEADER_LOOKUP.  When this value is found in `$_SERVER` (which aggregates ENV variables as well), the logic for identifying headers in the `$_SERVER` array will become strict.

For version 3.0, we will make the strict lookup the default.

Fixes #63

### DIFF
--- a/docs/book/v1/emitting-responses.md
+++ b/docs/book/v1/emitting-responses.md
@@ -1,6 +1,7 @@
 # Emitting Responses
 
-<!-- markdownlint-disable-next-line MD001 -->
+## Introduction
+
 > ### Deprecated
 >
 > Emitters are deprecated from Diactoros starting with version 1.8.0. The
--- a/docs/book/v1/emitting-responses.md
+++ b/docs/book/v1/emitting-responses.md
@@ -1,5 +1,6 @@
 # Emitting Responses
 
+<!-- markdownlint-disable-next-line MD001 -->
 > ### Deprecated
 >
 > Emitters are deprecated from Diactoros starting with version 1.8.0. The
--- a/src/functions/marshal_headers_from_sapi.php
+++ b/src/functions/marshal_headers_from_sapi.php
@@ -18,7 +18,7 @@ use function substr;
 function marshalHeadersFromSapi(array $server) : array
 {
     $contentHeaderLookup = isset($server['LAMINAS_DIACTOROS_STRICT_CONTENT_HEADER_LOOKUP'])
-        ? function (string $key) : bool {
+        ? static function (string $key) : bool {
             static $contentHeaders = [
                 'CONTENT_TYPE'   => true,
                 'CONTENT_LENGTH' => true,
@@ -26,7 +26,7 @@ function marshalHeadersFromSapi(array $server) : array
             ];
             return isset($contentHeaders[$key]);
         }
-        : function (string $key): bool {
+        : static function (string $key): bool {
             return strpos($key, 'CONTENT_') === 0;
         };
 
--- a/src/functions/marshal_headers_from_sapi.php
+++ b/src/functions/marshal_headers_from_sapi.php
@@ -17,6 +17,19 @@ use function substr;
  */
 function marshalHeadersFromSapi(array $server) : array
 {
+    $contentHeaderLookup = isset($server['LAMINAS_DIACTOROS_STRICT_CONTENT_HEADER_LOOKUP'])
+        ? function (string $key) : bool {
+            static $contentHeaders = [
+                'CONTENT_TYPE'   => true,
+                'CONTENT_LENGTH' => true,
+                'CONTENT_MD5'    => true,
+            ];
+            return isset($contentHeaders[$key]);
+        }
+        : function (string $key): bool {
+            return strpos($key, 'CONTENT_') === 0;
+        };
+
     $headers = [];
     foreach ($server as $key => $value) {
         if (! is_string($key)) {
@@ -45,7 +58,7 @@ function marshalHeadersFromSapi(array $server) : array
             continue;
         }
 
-        if (strpos($key, 'CONTENT_') === 0) {
+        if ($contentHeaderLookup($key)) {
             $name = strtr(strtolower($key), '_', '-');
             $headers[$name] = $value;
             continue;
--- a/docs/book/v2/usage.md
+++ b/docs/book/v2/usage.md
@@ -94,6 +94,27 @@ in the name were renamed with underlines. By getting the cookies directly from t
 access to the original cookies in the way you set them in your application and they are send by the user
 agent.
 
+> #### Strict Content- header matching
+>
+> Available since version 2.6.0
+>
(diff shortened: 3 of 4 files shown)


docs/book/v1/emitting-responses.md
```diff
@@ -1,5 +1,6 @@
 # Emitting Responses
 
+<!-- markdownlint-disable-next-line MD001 -->
```
weierophinney: I was getting a linting error due to this file, so fixed it here, as it was trivial, and easier than doing a separate PR.

src/functions/marshal_headers_from_sapi.php
```diff
@@ -17,6 +17,19 @@
  */
 function marshalHeadersFromSapi(array $server) : array
 {
+    $contentHeaderLookup = isset($server['LAMINAS_DIACTOROS_STRICT_CONTENT_HEADER_LOOKUP'])
+        ? function (string $key) : bool {
```
boesing: the function could be static and thus static variable could be omitted?

src/functions/marshal_headers_from_sapi.php
```diff
@@ -17,6 +17,19 @@
...
+            static $contentHeaders = [
+                'CONTENT_TYPE'   => true,
+                'CONTENT_LENGTH' => true,
+                'CONTENT_MD5'    => true,
+            ];
+            return isset($contentHeaders[$key]);
+        }
+        : function (string $key): bool {
```
boesing: maybe using static function for performance?

boesing (approved):
LGTM - maybe using static functions could lead to minor performance improvements but not 100% sure about this.

using a private constant could also be a way to get rid of the static variable.

docs/book/v1/emitting-responses.md
boesing: afaik that is due to the fact that there is no secondary header. maybe we should add one instead of ignoring the error?

bcremer:
Fix looks good for me :+1:

docs/book/v1/emitting-responses.md
weierophinney: No - this is a case where we have a "note", and those are indented at level 3.

src/functions/marshal_headers_from_sapi.php
weierophinney: I tried and got a linting error for the function definition, indicating static values cannot be dynamic. (I thought that would work, too, and had done it in the initial attempt.)

docs/book/v1/emitting-responses.md
froschdesign: > …and those are indented at level 3.

…and adding another header with level 2 before will solve the problem. 😉

src/functions/marshal_headers_from_sapi.php
weierophinney: PHP kills me sometimes. I can do it fine as part of the ternary, but when I was defining the functions as separate static variables previously, I was getting the linting error. Go figure.

weierophinney:
> using a private constant could also be a way to get rid of the static variable.

That would work if this were a class. However, since it's a function within a namespace, all constants are thus publicly visible.

docs/book/v1/emitting-responses.md
froschdesign:
```suggestion
## Introduction

```

weierophinney: We can do that in a different PR. :)

src/functions/marshal_headers_from_sapi.php
weierophinney: Also, static variables within the closure will help with performance as well, as they don't then need to be redefined on each invocation.

bcremer:
> For version 3.0, we will make the strict lookup the default.

I guess that never happened. The fix is still opt-in today for version 3.6: https://github.com/laminas/laminas-diactoros/blob/746db20f063a7cf11c6551c0a6d8aee7ddc7d236/src/functions/marshal_headers_from_sapi.php#L23

I will create a separate issue for that.

PR #69, opened 2021-08-17
Update PHPDoc: allow `StreamInterface` in `$streamOrFile` in `UploadedFile`

takoba:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description
When creating an UploadedFile that passed Stream, I was in trouble on PHPStan because `$streamOrFile` is not have annotation of StreamInterface. I improved it.

### DIFF
--- a/src/UploadedFile.php
+++ b/src/UploadedFile.php
@@ -80,7 +80,7 @@ class UploadedFile implements UploadedFileInterface
     private $stream;
 
     /**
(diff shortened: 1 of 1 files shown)


Ocramius (approved):
Thanks @takoba!

PR #72, opened 2021-09-03
Support for web servers that send HTTP2 request protocol as HTTP/2.0. Fixes #71

### LINKED ISSUE #71: Error with HTTP2 and NGINX
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 2.5.0

#### Summary

`Laminas\Diactoros\Exception\InvalidArgumentException: Unsupported HTTP protocol version "2.0" provided` exception when enabling http2 on nginx as it passes a protocol of `HTTP/2.0`

#### How to reproduce
Enable http2 on nginx (Current version tested is 1.21.1)

#### Expected behavior

No error with `HTTP/2.0`

JakeBooher:
|    Q          |   A
|-------------- | ------
| Documentation |no
| Bugfix        | yes
| BC Break      |no
| New Feature   |no
| RFC           |no
| QA            |no

### Description

Fixes #71

### DIFF
--- a/test/MessageTraitTest.php
+++ b/test/MessageTraitTest.php
@@ -50,7 +50,6 @@ class MessageTraitTest extends TestCase
             '1-without-minor'      => [ '1' ],
             '1-with-invalid-minor' => [ '1.2' ],
             '1-with-hotfix'        => [ '1.2.3' ],
-            '2-with-minor'         => [ '2.0' ],
         ];
     }
 
@@ -64,6 +63,27 @@ class MessageTraitTest extends TestCase
         $this->expectException(InvalidArgumentException::class);
 
         $request->withProtocolVersion($version);
+    }
+
+    public function validProtocolVersionProvider()
+    {
+        return [
+            '1.0' => [ '1.0' ],
+            '1.1' => [ '1.1' ],
+            '2'   => [ '2' ],
+            '2.0' => [ '2.0' ],
+        ];
+    }
+
+    /**
+     * @dataProvider validProtocolVersionProvider
+     */
+    public function testWithProtocolVersionDoesntRaiseExceptionForValidVersion($version)
+    {
+        $request = new Request();
+
+        $request->withProtocolVersion($version);
+        $this->addToAssertionCount(1);
     }
 
     public function testUsesStreamProvidedInConstructorAsBody()
--- a/test/MessageTraitTest.php
+++ b/test/MessageTraitTest.php
@@ -80,10 +80,8 @@ class MessageTraitTest extends TestCase
      */
     public function testWithProtocolVersionDoesntRaiseExceptionForValidVersion(string $version): void
     {
-        $request = new Request();
-
-        $request->withProtocolVersion($version);
-        $this->addToAssertionCount(1);
+        $request = (new Request())->withProtocolVersion($version);
+        $this->assertEquals($version, $request->getProtocolVersion());
     }
 
     public function testUsesStreamProvidedInConstructorAsBody(): void
--- a/test/MessageTraitTest.php
+++ b/test/MessageTraitTest.php
@@ -36,7 +36,7 @@ class MessageTraitTest extends TestCase
         $this->assertSame('1.0', $message->getProtocolVersion());
     }
 
-
+    /** @return non-empty-array<non-empty-string, array{mixed}> */
     public function invalidProtocolVersionProvider(): array
     {
         return [
@@ -55,6 +55,8 @@ class MessageTraitTest extends TestCase
 
     /**
      * @dataProvider invalidProtocolVersionProvider
+     *
+     * @param mixed $version
      */
     public function testWithProtocolVersionRaisesExceptionForInvalidVersion($version): void
     {
@@ -65,6 +67,7 @@ class MessageTraitTest extends TestCase
         $request->withProtocolVersion($version);
     }
 
+    /** @return non-empty-array<array{non-empty-string}> */
     public function validProtocolVersionProvider(): array
     {
         return [
@@ -179,6 +182,7 @@ class MessageTraitTest extends TestCase
         $this->assertSame(0, count($headers));
     }
 
+    /** @return non-empty-array<non-empty-string, array{mixed}> */
     public function invalidGeneralHeaderValues(): array
     {
         return [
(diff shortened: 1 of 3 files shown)


test/MessageTraitTest.php
```diff
@@ -50,7 +50,6 @@ public function invalidProtocolVersionProvider()
             '1-without-minor'      => [ '1' ],
             '1-with-invalid-minor' => [ '1.2' ],
             '1-with-hotfix'        => [ '1.2.3' ],
-            '2-with-minor'         => [ '2.0' ],
```
Ocramius, line 53: This used to be explicitly tested/marked invalid: did the RFC around it change?

Ocramius (changes requested):

test/MessageTraitTest.php
Xerkus: while protocol is called HTTP/2 we do not work with it here directly. And following is stated in the [ connection preface](https://www.rfc-editor.org/rfc/rfc7540.html#section-3.5) in the spec:

> In HTTP/2, each endpoint is required to send a connection preface as a final confirmation of the protocol in use and to establish the initial settings for the HTTP/2 connection.

> the connection preface starts with the string `PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n`

That implies that 2.0 is a correct value

boesing: if so, the comment should change as well, not?

Xerkus: I did not read any of the spec aside for quick scan. Does it matter outside of server request? `2.0` might need to be normalized to 2, i guess, but it should be accepted if (some) servers are sending it.

JakeBooher: This is a pretty much stock NGINX and PHP 7.4-fpm install, just added `http2` to the `listen` line and that's how it gets the value from `$_SERVER['SERVER_PROTOCOL']`

test/MessageTraitTest.php
```diff
@@ -66,6 +65,27 @@ public function testWithProtocolVersionRaisesExceptionForInvalidVersion($version
...
+
+    /**
+     * @dataProvider validProtocolVersionProvider
+     */
+    public function testWithProtocolVersionDoesntRaiseExceptionForValidVersion($version)
+    {
+        $request = new Request();
+
+        $request->withProtocolVersion($version);
+        $this->addToAssertionCount(1);
```
Ocramius: We should check the request protocol version at this point (instead of incrementing assertion counter)

test/MessageTraitTest.php
```diff
@@ -66,6 +65,27 @@ public function testWithProtocolVersionRaisesExceptionForInvalidVersion($version
...
+            '1.1' => [ '1.1' ],
+            '2'   => [ '2' ],
+            '2.0' => [ '2.0' ],
+        ];
+    }
+
+    /**
+     * @dataProvider validProtocolVersionProvider
+     */
+    public function testWithProtocolVersionDoesntRaiseExceptionForValidVersion($version)
```
Ocramius: Parameter/return type is to be provided

test/MessageTraitTest.php
```diff
@@ -66,6 +65,27 @@ public function testWithProtocolVersionRaisesExceptionForInvalidVersion($version
         $request->withProtocolVersion($version);
     }
 
+    public function validProtocolVersionProvider()
```
Ocramius: This is an `array<string, array{string}>`

Ocramius (approved):
Overall patch OK from my PoV, but needs some adjustments for it to be mergeable :)

test/MessageTraitTest.php
JakeBooher: These don't exist on any method within this class which is why I didn't add it. I'll go through and add to all the methods in this class for consistency if it's required for my changes

Ocramius:
Heyo, check the new type additions:

```
Error: Providers must return iterable<array-key, array<array-key, mixed>>, possibly different array<array-key, mixed> provided
Error: Providers must return iterable<array-key, array<array-key, mixed>>, possibly different array<array-key, mixed> provided
Error: Providers must return iterable<array-key, array<array-key, mixed>>, possibly different array<array-key, mixed> provided
Error: Providers must return iterable<array-key, array<array-key, mixed>>, possibly different array<array-key, mixed> provided
Error: Providers must return iterable<array-key, array<array-key, mixed>>, possibly different array<array-key, mixed> provided
Error: Providers must return iterable<array-key, array<array-key, mixed>>, possibly different array<array-key, mixed> provided
Error: Argument 2 of Psr\Http\Message\MessageInterface::withHeader expects array<array-key, string>|string, array{float} provided
Error: Argument 2 of Psr\Http\Message\MessageInterface::withHeader expects array<array-key, string>|string, float provided
Error: Providers must return iterable<array-key, array<array-key, mixed>>, possibly different array<array-key, mixed> provided
Error: Providers must return iterable<array-key, array<array-key, mixed>>, possibly different array<array-key, mixed> provided
```

That's because the provider type declarations should have `@return array<array<...>>` or such :)

JakeBooher:
Honestly have no time or interest in digging down that rabbit hole. Nothing even remotely resembling that exists anywhere in that test class, and I've already had to add return type hinting to every method in the class that didn't have it already. We can patch the original issue ourselves in a fork and include that instead if necessary. Feel free to close this PR and its associated issue if needed.

Ocramius:
I mean, if you think that's the responsible way to behave when trying to help on a project 🤷‍♀️

Sure, you can also just use a fork, but it really shows that you care little for being part of OSS, and are just hoping to reap the benefits.

We'll eventually get to fix it ourselves here, at some point.

test/MessageTraitTest.php
```diff
@@ -215,40 +233,40 @@ public function testWithHeaderReplacesDifferentCapitalization()
...
     }
 
-    public function testGetHeaderLineReturnsEmptyStringWhenHeaderDoesNotExist()
+    public function testGetHeaderLineReturnsEmptyStringWhenHeaderDoesNotExist(): void
     {
         $this->assertEmpty($this->message->getHeaderLine('X-Foo-Bar'));
     }
 
-    public function headersWithInjectionVectors()
+    public function headersWithInjectionVectors(): array
```
Ocramius, line 281: After local inspection, the reason why the psalm issues did arise is that all methods were given a new signature, instead of just those related to the fix: currently on it.

Ocramius:
I've adjusted the added type signatures accordingly - will now roll a release.

Thanks for the initial fix, @JakeBooher.

PR #80, opened 2022-02-21
Prepare for Renovate with reusable workflows

ghostwriter:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Adding Renovate to Laminas, Mezzio repositories. https://gist.github.com/weierophinney/49ca8ae2b916d898dd3d4162b26ff41b

- Configure minimum supported PHP version
- Update composer dependencies
- Update to reusable `Continuous Integration` workflow
- Update to reusable `Automatic Releases` workflow

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -4,30 +4,8 @@ on:
   pull_request:
   push:
     branches:
-      - '[0-9]+.[0-9]+.x'
-      - 'refs/pull/*'
     tags:
 
 jobs:
-  matrix:
-    name: Generate job matrix
-    runs-on: ubuntu-latest
-    outputs:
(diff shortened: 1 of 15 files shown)


Ocramius (changes requested):
Requires rolling back to a much smaller `composer.lock` change (just hash change)

weierophinney:
The failing issue for Psalm... I cannot reproduce locally, so I cannot add it to the psalm baseline. Basically, it's raising an error because of this line in `Stream.php`:

```php
        if (PHP_VERSION_ID >= 80000 && $resource instanceof \GdImage) {
```

`GdImage` was one of the resource pseudo-classes introduced in PHP 8.0, and `instanceof` checks _do not need for the class to exist_ for the check to run. :angry:

PR #82, opened 2022-04-06
Fix unserialization of header continuation / line folding

TimWolla:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

As per RFC 7230:

> A user agent that receives an obs-fold in a response message that is
> not within a message/http container MUST replace each received
> obs-fold with one or more SP octets prior to interpreting the field
> value.

Thus a space must be inserted when concatenating the continuation line in
AbstractSerializer::splitStream().

### DIFF
--- a/src/AbstractSerializer.php
+++ b/src/AbstractSerializer.php
@@ -109,7 +109,7 @@ abstract class AbstractSerializer
 
             // Append continuation to last header value found
             $value = array_pop($headers[$currentHeader]);
-            $headers[$currentHeader][] = $value . ltrim($line);
+            $headers[$currentHeader][] = $value . ' ' . ltrim($line);
         }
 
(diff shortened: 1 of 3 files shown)


Ocramius (approved):
LGTM - see https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.4

@weierophinney do we want this in a minor, or a patch? It's probably just a bug, so `2.9.x`?

Ocramius:
Targeting `2.9.1` and preparing a release, thanks @TimWolla!

weierophinney:
Yes - in a minor. Thanks, @Ocramius !

PR #87, opened 2022-05-04
Normalize line-folded headers values to not contain newlines

TimWolla:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | maybe?
| New Feature   | no
| RFC           | no
| QA            | no

### Description

As per RFC 7230#3.2.4:

> A server that receives an obs-fold in a request message that is not
> within a message/http container MUST […] replace
> each received obs-fold with one or more SP octets prior to
> interpreting the field value or forwarding the message downstream.
> […]
> A user agent that receives an obs-fold in a response message that is
> not within a message/http container MUST replace each received
> obs-fold with one or more SP octets prior to interpreting the field
> value.

Furthermore this change improves interoperability with PSR-7 implementations
that reject line folding.

The updated behavior matches the suggested behavior with regard to the handling
of line-folding in the new php-fig/fig-standards#1274 erratum.

### DIFF
--- a/src/MessageTrait.php
+++ b/src/MessageTrait.php
@@ -18,6 +18,7 @@ use function is_resource;
 use function is_string;
 use function preg_match;
 use function sprintf;
+use function str_replace;
 use function strtolower;
 
 /**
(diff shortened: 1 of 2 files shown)


Ocramius (approved):
LGTM. @weierophinney can you have a second look?

TimWolla:
Rebased to add the missing `use function`, see #88. I've also rebased onto 2.9.x to make it easier for you to pick into 2.9.x if you desire to do so.

Ocramius:
IMO OK for `2.10.x`

TimWolla:
> IMO OK for `2.10.x`

I'm fine either way. This is a fix for an obscure corner case of the HTTP spec and not something users will encounter in the real world.

weierophinney (approved):
:ship:

Thanks for the stellar errata for PSR-7, @TimWolla !

TimWolla:
> Thanks for the stellar errata for PSR-7, @TimWolla !

My pleasure. Thanks from my side for Diactoros and also for the super nice and speedy handling of my PRs :smiley:

PR #90, opened 2022-05-17
Use constant expression for JsonResponse::DEFAULT_JSON_FLAGS

TimWolla:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

Using a constant expression is legal in all PHP versions that are supported by Diactoros and prevents the expected values and actual values from diverging.

### DIFF
--- a/src/Response/JsonResponse.php
+++ b/src/Response/JsonResponse.php
@@ -16,6 +16,11 @@ use function json_last_error_msg;
 use function sprintf;
 
 use const JSON_ERROR_NONE;
+use const JSON_HEX_AMP;
+use const JSON_HEX_APOS;
+use const JSON_HEX_QUOT;
+use const JSON_HEX_TAG;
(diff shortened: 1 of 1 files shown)


Ocramius (approved):
Nice! Only thing left is the CS issue 😁

TimWolla:
> Nice! Only thing left is the CS issue grin

@Ocramius It's an exceeded line length, it's not obvious to me how that line should be wrapped. Do you have any suggestions?

Ocramius:
I'd try:

```php
$foo = BAR
    | BAZ;
```

If all else fails, we disable the rule.

TimWolla:
Thanks, updated. It's not particularly pretty, but passes phpcs locally.

Ocramius (approved):
LGTM, thanks @TimWolla!

PR #97, opened 2022-07-01
Ignore obviously malformed `host` headers when constructing a ServerRequest

TimWolla:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

The comment within the code should be self-explanatory. This adds some very basic checks to the host header as a hardening measure.

### DIFF
--- a/src/ServerRequestFactory.php
+++ b/src/ServerRequestFactory.php
@@ -168,7 +168,15 @@ class ServerRequestFactory implements ServerRequestFactoryInterface
 
         $host = self::getHeaderFromArray('host', $headers, false);
         if ($host !== false) {
-            return self::marshalHostAndPortFromHeader($host);
+            // Ignore obviously malformed host headers:
+            // - Whitespace is invalid within a hostname and break the URI representation within HTTP.
+            //   non-printable characters other than SPACE and TAB are already rejected by HeaderSecurity.
+            // - A comma indicates that multiple host headers have been sent which is not legal
+            //   and might be used in an attack where a load balancer sees a different host header
+            //   than Diactoros.
+            if (! \preg_match('/[\\t ,]/', $host)) {
+                return self::marshalHostAndPortFromHeader($host);
+            }
         }
 
         if (! isset($server['SERVER_NAME'])) {
--- a/src/ServerRequestFactory.php
+++ b/src/ServerRequestFactory.php
@@ -289,9 +289,10 @@ class ServerRequestFactory implements ServerRequestFactoryInterface
     /**
      * Retrieve a header value from an array of headers using a case-insensitive lookup.
      *
+     * @template T
      * @param array<string, string|list<string>> $headers Key/value header pairs
-     * @param mixed $default Default value to return if header not found
-     * @return mixed
+     * @param T $default Default value to return if header not found
+     * @return string|T
      */
     private static function getHeaderFromArray(string $name, array $headers, $default = null)
     {
--- a/test/ServerRequestFactoryTest.php
+++ b/test/ServerRequestFactoryTest.php
@@ -786,4 +786,40 @@ class ServerRequestFactoryTest extends TestCase
         $uri = $request->getUri();
         $this->assertSame('example.com', $uri->getHost());
     }
+
+    /**
+     * @psalm-return iterable<string, array{
+     *     0: string
+     * }>
+     */
+    public function invalidHostHeaders(): iterable
+    {
+        return [
+            'comma' => ['example.com,example.net'],
+            'space' => ['example com'],
+            'tab' => ["example\tcom"],
+        ];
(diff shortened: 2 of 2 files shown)


weierophinney (commented):
Quick question: does the `Uri` class not already disallow these values? Or is the change here primarily to just reject the `Host` header as a source for creating a `Uri` instance if it is malformed?

In other words, I'm wondering if we don't already get an exception that occurs for malformed `Host` headers.

TimWolla:
> Quick question: does the `Uri` class not already disallow these values?

No, the `Uri` class accepts arbitrary garbage as the `host`:

https://github.com/laminas/laminas-diactoros/blob/c272a93fc716456599d26bf7cc3281ccb708dabf/src/Uri.php#L297-L316

> In other words, I'm wondering if we don't already get an exception that occurs for malformed `Host` headers.

No, we don't. We likely should, but that would probably be a BC break. With this PR I wanted to add some minimally invasive checks to reject/ignore the most egregious violations that might actually cause harm, as the `host` header is possibly untrusted.

src/ServerRequestFactory.php
```diff
@@ -168,7 +168,15 @@ private static function marshalHostAndPort(array $server, array $headers) : arra
...
         $host = self::getHeaderFromArray('host', $headers, false);
         if ($host !== false) {
-            return self::marshalHostAndPortFromHeader($host);
+            // Ignore obviously malformed host headers:
+            // - Whitespace is invalid within a hostname and break the URI representation within HTTP.
+            //   non-printable characters other than SPACE and TAB are already rejected by HeaderSecurity.
+            // - A comma indicates that multiple host headers have been sent which is not legal
+            //   and might be used in an attack where a load balancer sees a different host header
+            //   than Diactoros.
+            if (! \preg_match('/[\\t ,]/', $host)) {
```
weierophinney, line 177:
```suggestion
            if (! \preg_match('/[\s,]/', $host)) {
```

`\s` will get spaces, tabs, and newlines.

test/ServerRequestFactoryTest.php
```diff
@@ -786,4 +786,40 @@ public function testHonorsHostHeaderOverServerNameWhenMarshalingUrl(): void
...
+     * @psalm-return iterable<string, array{
+     *     0: string
+     * }>
+     */
+    public function invalidHostHeaders(): iterable
+    {
+        return [
+            'comma' => ['example.com,example.net'],
+            'space' => ['example com'],
+            'tab' => ["example\tcom"],
```
weierophinney, line 800: Add another case here for entries containing newlines.

weierophinney (changes requested):

test/ServerRequestFactoryTest.php
TimWolla: These are already rejected by `HeaderSecurity` and thus the request completely fails to construct. See: https://github.com/laminas/laminas-diactoros/runs/7151517319?check_suite_focus=true

TimWolla: Should the `usage.md` be updated to explain that `ServerRequestFactory` may throw based on user input and any exceptions need to be gracefully handled: https://github.com/laminas/laminas-diactoros/blob/2.12.x/docs/book/v2/usage.md#marshaling-an-incoming-request?

see https://github.com/WoltLab/WCF/issues/4888

Ocramius:
Should this go against `2.11.x`, being a bugfix?

TimWolla:
> Should this go against 2.11.x, being a bugfix?

This technically has a (tiny) chance for breakage and is intended as a defense in depth measure. The web server / load balancer in front should already prevent those requests from reaching Diactoros. IMO 2.12 is fine.

src/ServerRequestFactory.php
```diff
@@ -289,9 +297,10 @@ private static function marshalHostAndPortFromHeader($host): array
     /**
      * Retrieve a header value from an array of headers using a case-insensitive lookup.
      *
+     * @template T
      * @param array<string, string|list<string>> $headers Key/value header pairs
-     * @param mixed $default Default value to return if header not found
-     * @return mixed
+     * @param T $default Default value to return if header not found
+     * @return string|T
```
Ocramius, line 303: Nice improvement!

test/ServerRequestFactoryTest.php
Ocramius: Not sure about it, but this discussion can be deferred to a different issue

Ocramius:
Rebased + manually merged in 54d4682ac51c005db15c50543e76796a86606fe9

That allowed shrinking of the baseline too.

Thanks @TimWolla!

PR #98, opened 2022-07-04
Fixed `UploadedFile::moveTo()` so it actually removes the original file when used in CLI context, and doesn't leave orphaned files

k2rn:
Signed-off-by: katsuren <katsuren.houken@gmail.com>
UploadedFile::moveTo doesn't work in CLI. Original file leave it as is.

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Calling UploadedFile::moveTo in CLI, it just copies the file to the targetPath.
In addition to this, the stream grabs the original file, so it cannot delete the file, and I cannot access the stream so I couldn't release the file pointer.
In a test for uploading file, I want to remove the uploaded file to keep clean the test folder.

### DIFF
--- a/src/UploadedFile.php
+++ b/src/UploadedFile.php
@@ -187,6 +187,13 @@ class UploadedFile implements UploadedFileInterface
             case (empty($sapi) || 0 === strpos($sapi, 'cli') || 0 === strpos($sapi, 'phpdbg') || ! $this->file):
                 // Non-SAPI environment, or no filename present
                 $this->writeFile($targetPath);
+
+                if ($this->stream instanceof StreamInterface) {
+                    $this->stream->close();
+                    if (is_string($this->file) && file_exists($this->file)) {
+                        unlink($this->file);
+                    }
+                }
                 break;
             default:
                 // SAPI environment, with file present
--- a/src/UploadedFile.php
+++ b/src/UploadedFile.php
@@ -190,9 +190,9 @@ class UploadedFile implements UploadedFileInterface
 
                 if ($this->stream instanceof StreamInterface) {
                     $this->stream->close();
-                    if (is_string($this->file) && file_exists($this->file)) {
-                        unlink($this->file);
-                    }
+                }
+                if (is_string($this->file) && file_exists($this->file)) {
+                    unlink($this->file);
                 }
                 break;
             default:
--- a/test/UploadedFileTest.php
+++ b/test/UploadedFileTest.php
@@ -32,17 +32,25 @@ use const UPLOAD_ERR_PARTIAL;
 
 class UploadedFileTest extends TestCase
 {
+    /** @var false|null|string */
+    protected $orgFile;
+
     protected $tmpFile;
 
     protected function setUp() : void
     {
-        $this->tmpfile = null;
+        $this->tmpFile = null;
+        $this->orgFile = null;
     }
 
     protected function tearDown() : void
     {
         if (is_string($this->tmpFile) && file_exists($this->tmpFile)) {
             unlink($this->tmpFile);
+        }
+
+        if (is_string($this->orgFile) && file_exists($this->orgFile)) {
+            unlink($this->orgFile);
         }
     }
 
@@ -142,17 +150,21 @@ class UploadedFileTest extends TestCase
         $this->assertSame($stream, $r->getValue($uploadStream));
     }
 
+    /**
+     * @return void
+     */
     public function testMovesFileToDesignatedPath()
     {
+        $originalContents = 'Foo bar!';
         $stream = new Stream('php://temp', 'wb+');
-        $stream->write('Foo bar!');
+        $stream->write($originalContents);
         $upload = new UploadedFile($stream, 0, UPLOAD_ERR_OK);
 
         $this->tmpFile = $to = tempnam(sys_get_temp_dir(), 'diac');
         $upload->moveTo($to);
         $this->assertTrue(file_exists($to));
         $contents = file_get_contents($to);
-        $this->assertSame($stream->__toString(), $contents);
+        $this->assertSame($originalContents, $contents);
     }
 
     public function invalidMovePaths()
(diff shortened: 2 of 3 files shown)


Ocramius (changes requested):
OK for me, but a test needs to be added here.

test/UploadedFileTest.php
```diff
@@ -146,13 +146,14 @@ public function testMovesFileToDesignatedPath()
     {
         $stream = new Stream('php://temp', 'wb+');
         $stream->write('Foo bar!');
+        $originalContents = $stream->__toString();
```
Ocramius: This leads to the stream being consumed before being passed to `UploadedFile` - this test should be left alone

test/UploadedFileTest.php
```diff
@@ -274,15 +275,17 @@ public function testGetStreamRaisesExceptionWhenErrorStatusPresent($status)
...
+        $uploadedFile = new UploadedFile($orgFile, 100, UPLOAD_ERR_OK, basename($orgFile), 'text/plain');
         $uploadedFile->moveTo($this->tmpFile);
 
-        $original = file_get_contents(__FILE__);
-        $test     = file_get_contents($this->tmpFile);
+        $contents = file_get_contents($this->tmpFile);
 
-        $this->assertSame($original, $test);
+        $this->assertSame($original, $contents);
     }
```
Ocramius: This test should be left as it was, to avoid hiding accidental BC breaks.

Ocramius (changes requested):

test/UploadedFileTest.php
k2rn: This test specifies `__FILE__` as resource.
It means UploadedFile::moveTo removes `__FILE__` after moving this TestCaseClass file.

Ocramius: Hmm, then we should not operate with `__FILE__`, since that is the test itself :thinking:

Ocramius: Can we put some contents in `$orgFile`?

test/UploadedFileTest.php
k2rn: Ah, I see. But `moveTo` function removes original stream.
So, I should save the contents before calling moveTo.

test/UploadedFileTest.php
k2rn: Should I replace all the `__FILE__` into `$this->orgFile` ?

Ocramius: A `"Hello"` written to it would be more than enough :-)

src/UploadedFile.php
```diff
@@ -187,6 +187,13 @@ public function moveTo($targetPath) : void
             case (empty($sapi) || 0 === strpos($sapi, 'cli') || 0 === strpos($sapi, 'phpdbg') || ! $this->file):
                 // Non-SAPI environment, or no filename present
                 $this->writeFile($targetPath);
+
+                if ($this->stream instanceof StreamInterface) {
+                    $this->stream->close();
+                    if (is_string($this->file) && file_exists($this->file)) {
+                        unlink($this->file);
+                    }
```
Ocramius: Should this cleanup happen regardless of the stream?

k2rn: In the `getStream` function, the resource is wrapped by StreamInterface in the end.
So maybe it is unnecessary to check the instance is StreamInterface.
Do you mean get the `if (is_string($this->file...` block out from `if ($this->stream ~` block?

Ocramius: Correct.

Ocramius:
Also to be asked here: do we merge to `2.11.x` (bugfix), or is it OK to merge to `2.12.x` (feature)? Would probably be better to release a bugfix.

k2rn:
I'm not in a hurry, so it's OK to `2.12.x` but, I'd appreciate if it could merge to `2.11.x`

test/UploadedFileTest.php
```diff
@@ -32,18 +32,26 @@
```
Ocramius, line 32: File mode change to be reverted

test/UploadedFileTest.php
```diff
@@ -320,4 +336,19 @@ public function testMoveToRaisesExceptionWithAppropriateMessageWhenUploadErrorDe
         $this->expectExceptionMessage($message);
         $uploadedFile->moveTo('/tmp/foo');
     }
+
+    /**
+     * @return void
```
Ocramius: Will declare these on the methods before merging

Ocramius (approved):

test/UploadedFileTest.php
Ocramius: 091487a175a4748aa7edb3f24f69f695f8bdb925

test/UploadedFileTest.php
Ocramius: 091487a175a4748aa7edb3f24f69f695f8bdb925

k2rn:
> @k2rn Please create an issue report to report the problem. Thanks in advance! 👍🏻

@froschdesign
https://github.com/laminas/laminas-diactoros/issues/108
I created the issue and close it. Thank you!

PowerKiKi:
I agree that this PR should be considered at the very least as a minor breaking change. It broke quite a few of my unit tests. At least it was easy to notice the breakage. But now I need to use what feels like a dirty workaround, something similar to:

```diff
 <?php

 use Laminas\Diactoros\Stream;
 use Laminas\Diactoros\UploadedFile;

- new UploadedFile('my-file-for-unit-test', 999, UPLOAD_ERR_OK),
+ new UploadedFile(new Stream('my-file-for-unit-test'), 999, UPLOAD_ERR_OK),
```

Ocramius:
Well, the fact that the file wasn't moved is an actual bug, and yes, it breaks compat for some tests, but unlikely for prod code. 

This is an XKCD 1172:

workflow_2x

PR #103, opened 2022-07-06
Update to `laminas/laminas-coding-standard:2.3.x`, improved types and internal API

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Does the donkey work upgrading to LCS 2.3

Changes to tests and src are in separate commits

### DIFF
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -105,9 +105,8 @@ class Stream implements StreamInterface
      * Attach a new stream/resource to the instance.
      *
      * @param string|resource $resource
-     * @throws Exception\InvalidArgumentException for stream identifier that cannot be
-     *     cast to a resource
-     * @throws Exception\InvalidArgumentException for non-resource stream
+     * @throws Exception\InvalidArgumentException For stream identifier that cannot be cast to a resource.
+     * @throws Exception\InvalidArgumentException For non-resource stream.
      */
     public function attach($resource, string $mode = 'r'): void
     {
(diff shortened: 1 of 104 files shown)


gsteel:
@Ocramius …
Integration tests are failing on 8.0 and 8.1 at https://github.com/laminas/laminas-diactoros/blob/c1a61118c23a8b996082e51a3c31a07f7f7107a5/src/Stream.php#L339

I think this is because an empty path is throwing a ValueError in `fopen` - In order to fix this failure, we'll need to explicitly check for the empty path and throw the expected exception right?

gsteel:
OK, I take that last comment back - for some reason, the error handler is being ignored on 8.0 and 8.1 - this is all stuff that hasn't been touched by this PR so I'm a little bit confused as to why CI would be failing here…

Ocramius (approved):
Awesome, thanks @gsteel!

Ocramius:
Ah, I did not notice the failures. I have another branch where I did this, and will propose a patch :thinking:

Ocramius:
Opened #104 - unsure if failures are comparable

src/Stream.php
```diff
@@ -320,23 +318,14 @@ public function getMetadata($key = null)
...
-            $resource = fopen($stream, $mode);
-            restore_error_handler();
-        }
-
-        if ($error) {
-            throw new Exception\RuntimeException('Invalid stream reference provided');
+            try {
+                $resource = fopen($stream, $mode);
+            } catch (Throwable $e) {
+                throw new Exception\RuntimeException('Invalid stream reference provided');
```
Ocramius: We should really rethrow the previous exception

Ocramius: s/rethrow/include

gsteel: Do you think it's a reasonable change - tests are now passing locally for me on 8.x
… The previous behaviour was to throw what's there - I'll amend to add previous perhaps? The integration tests would fail with anything other than a RuntimeException afaict

Ocramius: I think this is really only in case of a stream error - an exception should be rare, but is a healthy approach to a failure.

Ocramius (approved):
LGTM 🚢

PR #104, opened 2022-07-06
Upgraded codebase to comply with `laminas/laminas-coding-standard:2.3.x`

Ocramius:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

PR #105, opened 2022-07-06
Fix `symfony/error-handler` deprecations

andrew-demb:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes (?)
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

`symfony/error-handler` triggers deprecation errors in case of not (explicitly) specified return types when implemented interface have them declared (in phpdoc).
AFAIK they are triggered to make libraries migration to (possible) future declared return types more explicit.

It can be annoying to update all of libraries with such of changes, but these deprecation messages are pretty hard to silence :(

Here's log of there errors:
```
PHP Deprecated: Method "Psr\Http\Message\ServerRequestInterface::getParsedBody()" might add "array|object|null" as a native return type declaration in the future. Do the same in implementation "Laminas\Diactoros\ServerRequest" now to avoid errors or add an explicit @return annotation to suppress this message. in .../vendor/symfony/error-handler/DebugClassLoader.php on line 330
PHP Deprecated: Method "Psr\Http\Message\ServerRequestInterface::getAttribute()" might add "mixed" as a native return type declaration in the future. Do the same in implementation "Laminas\Diactoros\ServerRequest" now to avoid errors or add an explicit @return annotation to suppress this message. in .../vendor/symfony/error-handler/DebugClassLoader.php on line 330
PHP Deprecated: Method "Psr\Http\Message\StreamInterface::getMetadata()" might add "mixed" as a native return type declaration in the future. Do the same in implementation "Laminas\Diactoros\Stream" now to avoid errors or add an explicit @return annotation to suppress this message. in .../vendor/symfony/error-handler/DebugClassLoader.php on line 330
```

### DIFF
--- a/src/CallbackStream.php
+++ b/src/CallbackStream.php
@@ -159,6 +159,7 @@ class CallbackStream implements StreamInterface
 
     /**
      * {@inheritdoc}
+     * @return mixed
      */
     public function getMetadata($key = null)
     {
--- a/src/RelativeStream.php
+++ b/src/RelativeStream.php
@@ -169,6 +169,7 @@ final class RelativeStream implements StreamInterface
 
     /**
      * {@inheritdoc}
+     * @return mixed
      */
     public function getMetadata($key = null)
     {
--- a/src/ServerRequest.php
+++ b/src/ServerRequest.php
@@ -165,6 +165,7 @@ class ServerRequest implements ServerRequestInterface
 
     /**
      * {@inheritdoc}
+     * @return array|object|null
      */
     public function getParsedBody()
     {
@@ -199,6 +200,7 @@ class ServerRequest implements ServerRequestInterface
 
     /**
      * {@inheritdoc}
+     * @return mixed
      */
     public function getAttribute($attribute, $default = null)
     {
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -301,6 +301,7 @@ class Stream implements StreamInterface
 
     /**
      * {@inheritdoc}
+     * @return mixed
      */
     public function getMetadata($key = null)
     {


Ocramius:
Honestly, type improvements are welcome, but these are just messed up symfony issues.

andrew-demb:
Psalm fails the build because it cannot to read parameter type from interface, when there are phpdoc combines `@inheritdoc` and additional `@return` tag.

> Error: src/ServerRequest.php:205:34: MissingParamType: Parameter $attribute has no provided type (see https://psalm.dev/154)
> Error: src/ServerRequest.php:205:46: MissingParamType: Parameter $default has no provided type (see https://psalm.dev/154)

andrew-demb:
> Honestly, type improvements are welcome, but these are just messed up symfony issues.

Unfortunately I can't make types more precise for methods like `\Laminas\Diactoros\ServerRequest::getAttribute` [1].

Let's close then, we will try to hide errors from Symfony.

[1] https://github.com/laminas/laminas-diactoros/blob/c1a61118c23a8b996082e51a3c31a07f7f7107a5/src/ServerRequest.php#L200-L210

Ocramius:
My endorsement is to disable `symfony/error-handler` overall: it is a noxious component that effectively hooks in everything, worsening both runtime and developer experience.

The types refined in this particular patch were already precise enough in the interfaces declared by PSR-7, so IMO something else is happening.

Regardless of that, we are already working on improving types in #104 and #103.

Adding `@return mixed` on something where `{@inheritDoc}` already exists is really not improving on the situation, since `mixed` has little to no added value.

PR #106, opened 2022-07-07
Refined types as per `laminas/laminas-coding-standard:2.3.x` upgrades

Ocramius:
In https://github.com/laminas/laminas-diactoros/pull/104 (duplicate), we improved the types of internal and
public API. These were mixed with coding style improvements.

The coding style improvements have been merged separately via https://github.com/laminas/laminas-diactoros/pull/103,
so this patch only includes the type/docblock improvements, which should ease the potential evolution of the
API of this package.

|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### DIFF
--- a/src/Exception/InvalidStreamPointerPositionException.php
+++ b/src/Exception/InvalidStreamPointerPositionException.php
@@ -9,10 +9,10 @@ use Throwable;
 
 class InvalidStreamPointerPositionException extends RuntimeException implements ExceptionInterface
 {
-    /** @param int $code */
+    /** {@inheritDoc} */
     public function __construct(
         string $message = 'Invalid pointer position',
-        $code = 0,
+        int $code = 0,
         ?Throwable $previous = null
     ) {
         parent::__construct($message, $code, $previous);
--- a/src/Exception/UploadedFileAlreadyMovedException.php
+++ b/src/Exception/UploadedFileAlreadyMovedException.php
@@ -9,10 +9,10 @@ use Throwable;
 
 class UploadedFileAlreadyMovedException extends RuntimeException implements ExceptionInterface
 {
-    /** @param int $code */
+    /** {@inheritDoc} */
     public function __construct(
         string $message = 'Cannot retrieve stream after it has already moved',
-        $code = 0,
+        int $code = 0,
         ?Throwable $previous = null
     ) {
         parent::__construct($message, $code, $previous);
--- a/src/Request/ArraySerializer.php
+++ b/src/Request/ArraySerializer.php
@@ -23,6 +23,15 @@ final class ArraySerializer
 {
     /**
      * Serialize a request message to an array.
+     *
+     * @return array{
+     *     method: string,
+     *     request_target: string,
+     *     uri: string,
+     *     protocol_version: string,
+     *     headers: array<array<string>>,
+     *     body: string
+     * }
      */
     public static function toArray(RequestInterface $request): array
     {
--- a/src/Response/ArraySerializer.php
+++ b/src/Response/ArraySerializer.php
@@ -23,6 +23,14 @@ final class ArraySerializer
 {
     /**
      * Serialize a response message to an array.
+     *
+     * @return array{
+     *     status_code: int,
+     *     reason_phrase: string,
+     *     protocol_version: string,
+     *     headers: array<array<string>>,
+     *     body: string
+     * }
      */
     public static function toArray(ResponseInterface $response): array
     {
--- a/src/functions/parse_cookie_header.legacy.php
+++ b/src/functions/parse_cookie_header.legacy.php
@@ -11,7 +11,7 @@ use function Laminas\Diactoros\parseCookieHeader as laminas_parseCookieHeader;
  * @deprecated Use {@see \Laminas\Diactoros\parseCookieHeader} instead
  *
  * @param string $cookieHeader A string cookie header value.
- * @return array<string, string> key/value cookie pairs.
+ * @return array<non-empty-string, string> key/value cookie pairs.
  */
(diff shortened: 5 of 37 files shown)


Ocramius:
@gsteel you may be interested in checking this one - it's very closely related to your #104 work, but with some type refinements

src/Exception/InvalidStreamPointerPositionException.php
```diff
@@ -9,10 +9,10 @@
 
 class InvalidStreamPointerPositionException extends RuntimeException implements ExceptionInterface
 {
-    /** @param int $code */
+    /** {@inheritDoc} */
     public function __construct(
         string $message = 'Invalid pointer position',
-        $code = 0,
+        int $code = 0,
```
Ocramius, line 15: Non-BC break, as it is part of the constructor. Should a consumer pass a non-`int` here, due to `declare(strict_types=1)`, a crash would be happening during the `parent::__construct()` call anyway

src/Exception/UploadedFileAlreadyMovedException.php
```diff
@@ -9,10 +9,10 @@
 
 class UploadedFileAlreadyMovedException extends RuntimeException implements ExceptionInterface
 {
-    /** @param int $code */
+    /** {@inheritDoc} */
     public function __construct(
         string $message = 'Cannot retrieve stream after it has already moved',
-        $code = 0,
+        int $code = 0,
```
Ocramius, line 15: Non-BC break, as it is part of the constructor. Should a consumer pass a non-`int` here, due to `declare(strict_types=1)`, a crash would be happening during the `parent::__construct()` call anyway

test/HeaderSecurityTest.php
```diff
@@ -34,42 +34,47 @@ public function getFilterValues(): array
...
+            ["This is a \r\n\n\r\n test", false],
+            ["This is a \xFF test", false],
+            ["This is a \x7F test", false],
+            ["This is a \x7E test", true],
         ];
     }
 
     /**
      * @dataProvider validateValues
+     * @group ZF2015-04
```
Ocramius, line 69: Note: I re-introduced the various `@group` annotations.

While they may look outdated, they are really useful to dig up the reason why changes happened, especially after we moved code around repositories/subtree splits/etc.

gsteel: I'll remember that for next time 👍

src/Exception/UploadedFileAlreadyMovedException.php
boesing: ~We added `declare(strict_types=1)` in one of the latest minor. Wasn't that already a BC break then?~

nvm, it was there prior applying latest codestyle

src/Request/ArraySerializer.php
```diff
@@ -23,6 +23,15 @@ final class ArraySerializer
 {
     /**
      * Serialize a request message to an array.
+     *
+     * @return array{
```
boesing, line 27: Do we use this kind of array somewhere?
If so, I would prefer having a `psalm-type` so we can import that array shape for further usage.

src/Response/ArraySerializer.php
```diff
@@ -23,6 +23,14 @@ final class ArraySerializer
 {
     /**
      * Serialize a response message to an array.
+     *
+     * @return array{
```
boesing, line 27: Do we use this kind of array somewhere?
If so, I would prefer having a psalm-type so we can import that array shape for further usage.

src/functions/parse_cookie_header.php
```diff
@@ -16,7 +16,7 @@
  * overwriting. Thus, the server request should take the cookies from the request header instead.
  *
  * @param string $cookieHeader A string cookie header value.
- * @return array key/value cookie pairs.
+ * @return array<string, string> key/value cookie pairs.
```
boesing: can we be more strict here? guess its `array<non-empty-string,string>`?
I can't beliefe that a cookie header like this is supported:

```
Cookie: =
```

src/functions/parse_cookie_header.legacy.php
```diff
@@ -8,9 +8,10 @@
 use function Laminas\Diactoros\parseCookieHeader as laminas_parseCookieHeader;
```
boesing, line 8: Maybe we just drop this as we already removed support for the old zendframework implementation by adding `conflict` to `composer.json`?

boesing (changes requested):

src/Request/ArraySerializer.php
Ocramius: We don't currently use this

src/Response/ArraySerializer.php
Ocramius: We don't currently use this

src/functions/parse_cookie_header.php
Ocramius: Doesn't seem to be supported: will add `non-empty-string` here, although I cannot safely add an assertion, and will expand the baseline instead

src/functions/parse_cookie_header.legacy.php
Ocramius: We do declare these symbols: we can drop them, but not in this patch :thinking:

src/functions/parse_cookie_header.php
Ocramius: dfa990d294c2e6300699978e78c63724647c5d2a

gsteel (approved):

Ocramius:
:ship: then - will cut a release, hoping for quick feedback in case of breakages

PR #112, opened 2022-07-19
Add test to detect regression described in #111

### LINKED ISSUE #111: `FilterUsingXForwardedHeaders` should correctly deal with `<host>:<port>` pair in `X-FORWARDED-HOST` header
Hello contributors!

I get from my web-server headers like:

```
X-FORWARDED-HOST = host:port 
X-FORWARDED-PORT = port
```

I am using `FilterUsingXForwardedHeaders` filter and in result I get wierd URI from `ServerRequestFactory::fromGlobals` with `port` duplicated like `https://host:port:port/path`.

I guess`X-FORWARDED-HOST` in `FilterUsingXForwardedHeaders` should be parsed in the same way like `ServerRequestFactory::marshalHostAndPortFromHeader()` did.

In earlier versions (updated `2.9.2 -> 2.13.0`) it was working fine. So it's a kind of little `BC Break` too :)

I understand that I can implement `FilterServerRequestInterface` myself, but it would be more convenient to have such feature out of the box.

Thank you!

3DFace:
Signed-off-by: Denis Ponomarev <ponomarev@gmail.com>

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   |no
| RFC           | no
| QA            | yes

### Description

`FilterUsingXForwardedHeaders` should accept `host:port` pair in `X-FORWARDED-HOST` header. 

Test added to detect it parsed incorrectly.

### DIFF
--- a/test/ServerRequestFilter/FilterUsingXForwardedHeadersTest.php
+++ b/test/ServerRequestFilter/FilterUsingXForwardedHeadersTest.php
@@ -380,7 +380,7 @@ class FilterUsingXForwardedHeadersTest extends TestCase
             [
                 'Host'              => 'localhost',
                 'X-Forwarded-Host'  => 'example.com:4433',
-                'X-Forwarded-Port'  => '4433',
+                'X-Forwarded-Port'  => '8433',
                 'X-Forwarded-Proto' => 'https',
             ]
         );
--- a/test/ServerRequestFilter/FilterUsingXForwardedHeadersTest.php
+++ b/test/ServerRequestFilter/FilterUsingXForwardedHeadersTest.php
@@ -368,4 +368,30 @@ class FilterUsingXForwardedHeadersTest extends TestCase
         $uri             = $filteredRequest->getUri();
         $this->assertSame($expectedScheme, $uri->getScheme());
     }
+
+    public function testPortExtractedFromXForwardedHostHeader(): void
+    {
+        $request = new ServerRequest(
+            ['REMOTE_ADDR' => '192.168.1.1'],
+            [],
+            'http://localhost:80/foo/bar',
+            'GET',
+            'php://temp',
+            [
+                'Host'              => 'localhost',
+                'X-Forwarded-Host'  => 'example.com:4433',
+                'X-Forwarded-Port'  => '4433',
+                'X-Forwarded-Proto' => 'https',
+            ]
+        );
+
+        $filter = FilterUsingXForwardedHeaders::trustAny();
+
+        $filteredRequest = $filter($request);
+        $filteredUri     = $filteredRequest->getUri();
+        $this->assertNotSame($request->getUri(), $filteredUri);
+        $this->assertSame('example.com', $filteredUri->getHost());
+        $this->assertSame(4433, $filteredUri->getPort());
+        $this->assertSame('https', $filteredUri->getScheme());
+    }
 }
--- a/.gitattributes
+++ b/.gitattributes
@@ -7,4 +7,5 @@
 /mkdocs.yml export-ignore
 /phpcs.xml export-ignore
 /phpunit.xml.dist export-ignore
+/renovate.json export-ignore
 /test/ export-ignore
--- a/composer.json
+++ b/composer.json
@@ -25,9 +25,6 @@
         "sort-packages": true,
         "platform": {
             "php": "7.3.99"
-        },
-        "allow-plugins": {
-            "dealerdirect/phpcodesniffer-composer-installer": true
         }
     },
     "extra": {


test/ServerRequestFilter/FilterUsingXForwardedHeadersTest.php
```diff
@@ -368,4 +368,30 @@ public function testOnlyHonorsXForwardedProtoIfValueResolvesToHTTPS(
...
+    {
+        $request = new ServerRequest(
+            ['REMOTE_ADDR' => '192.168.1.1'],
+            [],
+            'http://localhost:80/foo/bar',
+            'GET',
+            'php://temp',
+            [
+                'Host'              => 'localhost',
+                'X-Forwarded-Host'  => 'example.com:4433',
```
Ocramius, line 382: We should probably comment on which kind of server/proxy does this :thinking:

test/ServerRequestFilter/FilterUsingXForwardedHeadersTest.php
```diff
@@ -368,4 +368,30 @@ public function testOnlyHonorsXForwardedProtoIfValueResolvesToHTTPS(
...
+                'X-Forwarded-Port'  => '4433',
```
Ocramius: I think `X-Forwarded-Port` should be different, to make sure that the header contents take priority over `X-Forwarded-Host`

Ocramius (changes requested):

test/ServerRequestFilter/FilterUsingXForwardedHeadersTest.php
3DFace: It's difficult to me to investigate how exactly `X-FORWARDED-HOST` is made.
In my case `php-fpm` is behind `nginx` which iteself is behind `traefik`. I can't tell which one did that...

test/ServerRequestFilter/FilterUsingXForwardedHeadersTest.php
3DFace:
```suggestion
                'X-Forwarded-Port'  => '8433',
```

test/ServerRequestFilter/FilterUsingXForwardedHeadersTest.php
Ocramius: Are you able to log a full request coming from traefik (in nginx)? Perhaps that will tell you something more.

Practically, we want to give this test a valid name, that explains whether this is Traefik **or** Nginx.

3DFace: I did logging - it comes from **Traefik**.

But I'm not sure it should be used to name a test method though. I guess there are more possible servers that can do the same...
Example at wikipedia shows `host:port` format as possible:
https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#Common_non-standard_request_fields

Ocramius:
Ok, now the question is just what to do with it inside the marshalling logic :D

test/ServerRequestFilter/FilterUsingXForwardedHeadersTest.php
```diff
@@ -368,4 +368,30 @@ public function testOnlyHonorsXForwardedProtoIfValueResolvesToHTTPS(
...
+            ]
+        );
+
+        $filter = FilterUsingXForwardedHeaders::trustAny();
+
+        $filteredRequest = $filter($request);
+        $filteredUri     = $filteredRequest->getUri();
+        $this->assertNotSame($request->getUri(), $filteredUri);
+        $this->assertSame('example.com', $filteredUri->getHost());
+        $this->assertSame(4433, $filteredUri->getPort());
```
boesing, line 394: What port had priority prior the security fix?
Was that the port from the `X-Forwarded-Host` or `X-Forwarded-Port`?

boesing:
> Ok, now the question is just what to do with it inside the marshalling logic :D

I'd say we should provide the pre-fix solution. I can not have a look into it right now but lets see what the outcome of my [question](https://github.com/laminas/laminas-diactoros/pull/112#discussion_r924470179) is

test/ServerRequestFilter/FilterUsingXForwardedHeadersTest.php
3DFace: It seems like `x-forwarded-port` was not checked at all - did not found it in 2.9.2.
```
$forwardedHost = $getHeaderFromArray('x-forwarded-host', $headers, false);
if ($forwardedHost !== false) {
    return $marshalHostAndPortFromHeader($forwardedHost);
}

$host = $getHeaderFromArray('host', $headers, false);
if ($host !== false) {
    return $marshalHostAndPortFromHeader($host);
}
```

So `x-forwarded-host` and `host` was treated in the same way.

3DFace:
BTW this is my first ever pull-request. I will be grateful for explicit instructions if any actions are expected from me.

Ocramius:
@3DFace so far, all is correct, now we just need to fix this :D 

If you have a clear idea of what is needed, you can just push new commits to your branch (with the proposed changes), otherwise somebody else will take your commits and propose a patch.

If you have time and capacity, then just give it a shot: change the sources that affect this test outcome, until the tests are all green again.

Ocramius:
As for the other 5 unrelated commits: I will get rid of them before merging, or they can be rebased out: that's because this is a bugfix, so it should go to `2.13.x` (next patch release branch), and your patch started from the default branch (`2.14.x`).

That's not a big problem though, and can be addressed later.

3DFace:
> As for the other 5 unrelated commits

Thank you for clarifications! I was a bit confused about it.

test/ServerRequestFilter/FilterUsingXForwardedHeadersTest.php
weierophinney: Then maybe something along the lines of `testPortCanBeExtractedFromXForwardedHostHeaderWhenPresent()`.

test/ServerRequestFilter/FilterUsingXForwardedHeadersTest.php
weierophinney: Prior to the fix, we didn't consider `X-Forwarded-Port`, nor did we consider `X-Forwarded-Host` values with ports included; in the latter case, this likely would have led to an exception thrown by the `Uri` instance during creation, as the host value would fail assertions.

Since we honor `X-Forwarded-Port` now, I think it makes sense to provide it, and it's technically a bug in even the 2.9.2 implementation.

weierophinney (commented):
I wouldn't call this a regression, as we never supported it. The fact that we can now (due to the fact that `X-Forwarded-Port` is now honored) would be technically a new feature; the fact that if we were properly following the spec, it would have been present all along means it could be considered a bug.

In terms of priority, I'd point to the [Forwarded header spec](https://datatracker.ietf.org/doc/html/rfc7239), and the fact that it **does not** include a separate key for port; if a port is specified, it's part of the "host" key. As such, I'd recommend that if an `X-Forwarded-Port` header is present AND the `X-Forwarded-Host` header _also_ includes a port, we would either:

- raise an exception
- prefer the port in the `X-Forwarded-Host` header (and document what we do) AND trigger a warning

The first would be most correct, but could lead to a DoS (as a malicious proxy could send a flood of requests with conflicting headers that then lead to the dreaded "white screen"). The second would prevent DoS scenarios, but would be harder to detect (though good monitoring rules would help identify them).

boesing:
I've created a fix in #111 

Due to the fact that at least apache is blindly passing `Host` header via `X-Forwarded-Host` and the `Host` header allows containing a port (even tho it won't do anything), I'd say the correct behavior would be to simply ignore the port from `X-Forwarded-Host`.

If the behavior should be that we are able to determine the initial request / port, we shouldn't use `X-Forwarded-Host` or `X-Forwarded-Port` as thats always those values from the latest proxy (at least that is what I would assume, not sure if proxies do actually pass these values in case they're existing).
Instead, we might want to use `Forwarded` which contains the whole history (in correct order) of all proxies which interacted with the request containing the hostnames they received. But that would be a whole another topic.

In the OP scenario, that would be traefik. We are only able to determine those headers from NGINX which are most probably not those we might want to use here.

weierophinney:
@boesing Should we close this in favor of the solution in #135?

boesing:
@weierophinney would be fine for me, I guess the issue is covered?

PR #113, opened 2022-07-28
Fix static return type in RequestTrait

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| QA            | yes

### Description

Solves psalm issues here and for consumers where a call to `ServerRequest::withWhatever()` would be inferred as `RequestInterface` rather than `ServerRequestInterface`


Ocramius (approved):
LGTM :+1:

PR #115, opened 2022-08-28
Apply PHP 7.4 syntax and typed property

samsonasik:
Signed-off-by: Abdul Malik Ikhsan <samsonasik@gmail.com>

|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Since composer.json require php 7.4, php 7.4 syntax can be applied with typed properties. 

- For `final class`, typed properties for protected is allowed if no inherit of parent or used by its final class parent.
- Otherwise, update only private properties.

### DIFF
--- a/src/functions/marshal_uri_from_sapi.php
+++ b/src/functions/marshal_uri_from_sapi.php
@@ -42,7 +42,7 @@ function marshalUriFromSapi(array $server, array $headers): Uri
      * @param mixed $default Default value to return if header not found
      * @return mixed
      */
-    $getHeaderFromArray = function (string $name, array $headers, $default = null) {
+    $getHeaderFromArray = static function (string $name, array $headers, $default = null) {
         $header  = strtolower($name);
         $headers = array_change_key_case($headers, CASE_LOWER);
         if (array_key_exists($header, $headers)) {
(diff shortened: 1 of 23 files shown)


src/functions/marshal_uri_from_sapi.php
```diff
@@ -138,45 +126,39 @@ function marshalUriFromSapi(array $server, array $headers): Uri
...
     $uri = new Uri('');
 
     // URI scheme
     $scheme            = 'http';
-    $marshalHttpsValue = function ($https): bool {
+    $marshalHttpsValue = static function ($https): bool {
         if (is_bool($https)) {
             return $https;
         }
-
```
Ocramius: Not sure why your CS adjustments removed all these spaces: seems invalid?

Makes the entire thing much less readable too...

Ocramius (changes requested):

src/functions/marshal_uri_from_sapi.php
samsonasik: Yes, that's known drawback of using rector https://github.com/rectorphp/rector#known-drawbacks , I can rollback the lines if needed.

Ocramius: Yes please: I know it's annoying, but some of these spacing changes do need adjustment indeed :)

samsonasik: I rolled back spaces https://github.com/laminas/laminas-diactoros/pull/115/commits/cf8ec668d3f7c189fb578a32992f488914b3603a

samsonasik:
@Ocramius All green 🎉 it is ready now.

Ocramius (approved):
Thanks @samsonasik!

PR #116, opened 2022-08-30
Apply PHP 8.2's `SensitiveParameter` attribute to `Uri::withUserInfo()`

TimWolla:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | maybe

### Description

Technically the `$password` will end up in the URI anyway when stringifying it.

Adding the Attribute is simple though and absolves the reader from needing to
consider whether not having the attribute in this specific instance is safe or
not.

-------------------

I've searched the codebase for “password“, “auth” and “secret” and this is the only place where the attribute makes sense.

### DIFF
--- a/src/Uri.php
+++ b/src/Uri.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Laminas\Diactoros;
 
 use Psr\Http\Message\UriInterface;
+use SensitiveParameter;
 
 use function array_keys;
 use function explode;
@@ -228,6 +229,9 @@ class Uri implements UriInterface
         return $new;
     }
 
+    // The following rule is buggy for parameters attributes
+    // phpcs:disable SlevomatCodingStandard.TypeHints.ParameterTypeHintSpacing.NoSpaceBetweenTypeHintAndParameter
+
     /**
      * Create and return a new instance containing the provided user credentials.
      *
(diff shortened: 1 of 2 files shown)


TimWolla:
> Error: There must be exactly one space between parameter type hint and parameter $password.

That likely is a bug in PHP CodeSniffer?

src/Uri.php
```diff
@@ -236,8 +237,11 @@ public function withScheme($scheme): UriInterface
      *
      * {@inheritdoc}
      */
-    public function withUserInfo($user, $password = null): UriInterface
-    {
+    public function withUserInfo(
+        $user,
+        #[SensitiveParameter]
```
Ocramius, line 245: This does indeed look like a bug in PHPCS.

Disabling the rule for this file could suffice.

Ocramius (changes requested):

psalm.xml.dist
```diff
@@ -43,6 +43,17 @@
                 <referencedClass name="GdImage" />
             </errorLevel>
         </UndefinedClass>
+
+        <UndefinedAttributeClass>
```
Ocramius: I would say that we add this to the baseline: this is not something we explicitly ignore, but rather will disappear with the PHP version upgrade

Ocramius (approved):
Thanks @TimWolla!

Issue #122, opened 2022-11-14
Could ServerRequestFactory::marshallUriFromSapi() be made public?


markstory:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | yes

#### Summary

:wave: from your friends at the CakePHP project, we've been longtime users of diactoros :bow: We recently learned of the plans to deprecate and remove `marshalUriFromSapi` and eventually remove it.

https://github.com/laminas/laminas-diactoros/blob/e5b6419fea007b8b4a71034edc8ec96c88d4c57d/src/functions/marshal_uri_from_sapi.php#L30-L36

The other implementation (`ServerRequestFactory::marshallUriFromSapi()`) is currently private. 

https://github.com/laminas/laminas-diactoros/blob/e5b6419fea007b8b4a71034edc8ec96c88d4c57d/src/ServerRequestFactory.php#L124

Is there any appetite in making this method public so that we could use it in our [RequestFactory](https://github.com/cakephp/cakephp/blob/a5f9900b1bdc652dcafc64b525587d54a6b8d48c/src/Http/ServerRequestFactory.php#L226-L233)?

Ocramius:
Depends on the use-case: it would need to be extracted from `ServerRequestFactory`, IMO.

Wouldn't using this work? 

https://github.com/laminas/laminas-diactoros/blob/0b7d4e8ff6560e3a7f84d16283f2e12402f2bd78/src/ServerRequestFactory.php#L67

Can you build a request and then get the URI from there?

ADmad:
> Can you build a request and then get the URI from there?

We build our own `ServerRequest` instance (using the various global functions Diactoros provides), so building Diactoros' `ServerRequest` instance just to get the `Uri` instance would be quite wasteful. All the parsing of PHP's global environment/request  variables is non trivial amount of code, so would like to avoid doing it twice.

Ocramius:
Well aware, it's just that the public API becomes much bigger there.

If somebody can send me a patch to try and extract `marshalUriFromSapi()` into a well documented `public` API on a new class that we test independently (and can deprecate separately, if needed in future), that would probably be the best way forward :)

Ocramius:
Handled in #123

PR #123, opened 2022-11-21
Add `UriFactory::createFromSapi()`

markstory:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | yes
| QA            | no

### Description

This method will provide a public API to create Uri instances from SERVER context. This helps libraries that depend on diactoros and have their own ServerRequest or Uri implementations use the logic in diactoros more efficiently.

I wanted to get a rough draft of this up for feedback. If it is headed in the right direction, I have a few TODOs remaining:

* [ ] Complete the tests (coverage isn't quite on par with the rest of the library).
* [ ] Write documentation. I was thinking of putting it the `factories` section, but would appreciate any direction here.

If there are other requirements that need to be met let me know :smile:

Refs #122


src/UriFactory.php
```diff
@@ -16,4 +36,217 @@ public function createUri(string $uri = ''): UriInterface
     {
         return new Uri($uri);
     }
+
+    /**
+     * Create a Uri instance based on the headers and $_SERVER data.
+     *
+     * @param array<string, string|list<string>> $headers
+     * @param array $server SAPI parameters
```
Ocramius: Psalm has a quite decent coverage of `$_SERVER`: https://github.com/vimeo/psalm/blob/7869cb548a5d88399f971ce9f897fe83abbd4213/src/Psalm/Internal/Analyzer/Statements/Expression/Fetch/VariableFetchAnalyzer.php#L671

Seems like it's an `array<non-empty-string, list<string>|int|float|string>>` from there, but I'm a bit groggy, so it might be worth double-checking

src/UriFactory.php
```diff
@@ -16,4 +36,217 @@ public function createUri(string $uri = ''): UriInterface
...
+     */
+    public static function createFromSapi(array $server, array $headers): Uri
+    {
+        $uri = new Uri('');
+
+        // URI scheme
```
Ocramius: These comments are kinda redundant with the code :D 

Perhaps better variable names could replace them?

src/UriFactory.php
```diff
@@ -16,4 +36,217 @@ public function createUri(string $uri = ''): UriInterface
...
+        $host = self::getHeaderFromArray('host', $headers, false);
+        if ($host !== false) {
+            // Ignore obviously malformed host headers:
+            // - Whitespace is invalid within a hostname and break the URI representation within HTTP.
+            //   non-printable characters other than SPACE and TAB are already rejected by HeaderSecurity.
+            // - A comma indicates that multiple host headers have been sent which is not legal
+            //   and might be used in an attack where a load balancer sees a different host header
+            //   than Diactoros.
+            if (! preg_match('/[\\t ,]/', $host)) {
+                return self::marshalHostAndPortFromHeader($host);
```
Ocramius, line 132: Some of these type check issues require careful checking: in this specific call, for example, `marshalHostAndPortFromHeader` could get better declared types.

test/UriFactoryTest.php
```diff
@@ -0,0 +1,256 @@
...
+use PHPUnit\Framework\TestCase;
+
+use function array_shift;
+use function sprintf;
+use function str_contains;
+use function strtolower;
+
+class UriFactoryTest extends TestCase
+{
+    public function testCreateFromSapiUsesIISUnencodedUrlValueIfPresentAndUrlWasRewritten(): void
```
Ocramius, line 18: Did we have existing test coverage for `private static function marshalUriFromSapi(`? Perhaps they can just be ported here...

Ocramius (changes requested):

Ocramius:
BTW, overall direction of this patch is good :+1:

test/UriFactoryTest.php
markstory: Yes, the bulk of these tests were extracted from the `ServerRequestFactory` tests.

markstory:
I've resolved all but one of psalm's complaints. The remaining MixedArgumentTypeCoercion was added to the baseline file beside the existing one in the same method.

Ocramius (approved):
LGTM 👍

Thanks @markstory!

Ocramius:
Went through it again: I know most code is just transposed as-is, so we're good here.

Releasing 🚢

PR #124, opened 2022-11-21
Return static from withStatus

adamturcsan:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | yes
| New Feature   | no
| RFC           | yes
| QA            | no

### Description

As this library requires at least php 8.0 it now can make use of the `static` return type.

PSR RepsonseInterface [states](https://www.php-fig.org/psr/psr-7/#33-psrhttpmessageresponseinterface) that a withStatus method should return `static`.

Now `Response` as a return type can meddle with decoration type informations in static analyzers.

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -144,7 +144,7 @@ class Response implements ResponseInterface
     /**
      * {@inheritdoc}
      */
-    public function withStatus($code, $reasonPhrase = ''): Response
+    public function withStatus($code, $reasonPhrase = ''): static
     {
         $new = clone $this;
         $new->setStatusCode($code, $reasonPhrase);


internalsystemerror (commented):
If there is a class already extending this with Response as their signature, would this change require them to update their child class?

src/Response.php
```diff
@@ -144,7 +144,7 @@ public function getReasonPhrase(): string
     /**
      * {@inheritdoc}
      */
-    public function withStatus($code, $reasonPhrase = ''): Response
+    public function withStatus($code, $reasonPhrase = ''): static
     {
```
Ocramius, line 148: This 100% breaks inheritance: can only be done at docblock level

Ocramius (changes requested):

src/Response.php
adamturcsan: Docblock already states `static` as it is inherited from PSR ResponseInterface. The issue is that the code (the function declartion) doesn't comply with that so some info gets lost after a `withStatus`.
By inspecting the implementation we see it in fact returns `static` and it is an issue only for some static analysis.

I see the BC break though, but for now I don't see how I could provide further docblock info.

Ocramius: If the docblock already states `@return static`, then this patch shouldn't be necessary for static analysis purposes

adamturcsan: It may be an issue in phpstan then.
Let me provide a wider context with a code example, and then I'll shut up :smiley:  and search for a solution from that POV.

```
use Psr\Http\Message\ResponseInterface;
use Laminas\Diactoros\Response\JsonResponse;

class DecoratedResponse implements ResponseInterface {
    private JsonResponse $jsonResponse;

    /* A lot of other method delegation to $this->jsonResponse */

    public function withStatus($code, $reasonPhrase = ''): static
    {
        $newInstance = clone $this;
        $newInstance->jsonResponse = $this->jsonResponse->withStatus($code, $reasonPhrase);
        return $newInstance;
    }
}
```
This delegation doesn't return the expected type:

`DecoratedResponse::$jsonResponse (Laminas\Diactoros\Response\JsonResponse) does not accept Laminas\Diactoros\Response.`

Ocramius: Can you perhaps try changing `{@inheritDoc}` to `@inheritDoc`? Perhaps phpstan is not prioritizing the docblock.

adamturcsan: It didn't do much. I changed it to `@return static` then it worked. So I guess phpstan doesn't handle inheriting type information from docblock well.

Ocramius:
Closing here as per @adamturcsan's findings: we cannot change the native type here, without breaking downstream BC, and PHPStan seems to prioritise native type to docblock type.

My endorsement is to look for (or raise) an issue in `phpstan/phpstan-src`

PR #128, opened 2022-12-14
Update PSR-7 integration tests

### LINKED ISSUE #74: PSR-7 Integration Tests Failing

weierophinney:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | slight behavior change
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

Updates to 1.2.0 of the PSR-7 integration test suite, which adds the tests we wrote for mitigating ZF2015-05, and requires a few changes:

- When creating the string representation of the URL, we DO NOT normalize the path to remove multiple leading slashes. In its absolute form, this is not necessary.
- All normalization is done via `getPath()`; this mitigates the common XSS scenario.
- It adds a test to validate that when using origin-form during a `RequestInterface::getRequestTarget()` call, it will use the results of `getPath()`, as this is a scenario where the XSS could also occur.

I have removed one test from `UriTest`, as it contradicts the first point above. Since the scenario is covered in the PSR-7 integration tests, we are covered.

See php-http/psr7-integration-tests#54 for more details.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,7 @@
         "ext-libxml": "*",
         "http-interop/http-factory-tests": "^0.9.0",
         "laminas/laminas-coding-standard": "^2.4.0",
-        "php-http/psr7-integration-tests": "^1.1.1",
+        "php-http/psr7-integration-tests": "^1.2",
         "phpunit/phpunit": "^9.5.26",
         "psalm/plugin-phpunit": "^0.18.0",
         "vimeo/psalm": "^5.0.0"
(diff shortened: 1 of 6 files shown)


l0gicgate (approved):

Ocramius (approved):
Excellent, thanks @weierophinney!

PR #133, opened 2023-04-04
Allow usage with PSR-7 v1.1

weierophinney:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | yes

### Description

This patch updates the PSR-7 minimum supported version to 1.1. PSR-7 1.1 adds parameter type declarations, but not return type declarations. Since parameters can be widened in implementations, and return types narrowed, our existing implementation (which already features return types) already complies.

For testing, I've pinned this to a branch of the PSR-7 integration test suite submitted as php-http/psr7-integration-tests#68. This ensures that we fulfill the changed expectations (raising either `InvalidArgumentException` OR `TypeError` for invalid parameters), and will not need to update again once we prepare a v3 release of this library that explicitly adds parameter type declarations.

After updating dependencies, Psalm flagged a number of issues. I corrected one that was easily corrected; the remainder are due to Psalm detecting the parent parameter type declarations and deciding we don't need to do some of our validations. Since at runtime, invalid data could still be passed to the implementation, I've pushed these into the Psalm baseline for now, and for removal once we prepare for v3 targeting PSR-7 v2.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
     "require": {
         "php": "~8.0.0 || ~8.1.0 || ~8.2.0",
         "psr/http-factory": "^1.0",
-        "psr/http-message": "^1.0"
+        "psr/http-message": "^1.1"
     },
     "require-dev": {
         "ext-curl": "*",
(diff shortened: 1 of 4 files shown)


gsteel (approved):
LGTM!

boesing (approved):
LGTM

PR #135, opened 2023-04-05
Examine `X-Forwarded-Host` for additional port information

### LINKED ISSUE #111: `FilterUsingXForwardedHeaders` should correctly deal with `<host>:<port>` pair in `X-FORWARDED-HOST` header

boesing:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes

### Description

Fixes #111

### DIFF
--- a/src/ServerRequestFilter/FilterUsingXForwardedHeaders.php
+++ b/src/ServerRequestFilter/FilterUsingXForwardedHeaders.php
@@ -81,7 +81,7 @@ final class FilterUsingXForwardedHeaders implements FilterServerRequestInterface
                     $uri           = $uri
                         ->withHost($host);
                     if ($port !== null) {
-                        $uri = $uri->withPort((int) $port);
+                        $uri = $uri->withPort($port);
                     }
                     break;
                 case self::HEADER_PORT:
--- a/src/ServerRequestFilter/FilterUsingXForwardedHeaders.php
+++ b/src/ServerRequestFilter/FilterUsingXForwardedHeaders.php
@@ -6,6 +6,7 @@ namespace Laminas\Diactoros\ServerRequestFilter;
 
 use Laminas\Diactoros\Exception\InvalidForwardedHeaderNameException;
 use Laminas\Diactoros\Exception\InvalidProxyAddressException;
+use Laminas\Diactoros\UriFactory;
 use Psr\Http\Message\ServerRequestInterface;
 
 use function explode;
@@ -76,7 +77,8 @@ final class FilterUsingXForwardedHeaders implements FilterServerRequestInterface
 
             switch ($headerName) {
                 case self::HEADER_HOST:
-                    $uri = $uri->withHost($header);
+                    [$host] = UriFactory::marshalHostAndPortFromHeader($header);
+                    $uri    = $uri->withHost($host);
                     break;
                 case self::HEADER_PORT:
                     $uri = $uri->withPort((int) $header);
--- a/src/ServerRequestFilter/FilterUsingXForwardedHeaders.php
+++ b/src/ServerRequestFilter/FilterUsingXForwardedHeaders.php
@@ -77,8 +77,12 @@ final class FilterUsingXForwardedHeaders implements FilterServerRequestInterface
 
             switch ($headerName) {
                 case self::HEADER_HOST:
-                    [$host] = UriFactory::marshalHostAndPortFromHeader($header);
-                    $uri    = $uri->withHost($host);
+                    [$host, $port] = UriFactory::marshalHostAndPortFromHeader($header);
+                    $uri           = $uri
+                        ->withHost($host);
+                    if ($port !== null) {
+                        $uri = $uri->withPort((int) $port);
+                    }
                     break;
                 case self::HEADER_PORT:
                     $uri = $uri->withPort((int) $header);
--- a/.gitattributes
+++ b/.gitattributes
@@ -12,3 +12,4 @@
 /psalm.xml.dist export-ignore
 /renovate.json export-ignore
 /test/ export-ignore
+/psalm/ export-ignore
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="5.3.0@b6faa3e96b8eb50ec71384c53799b8a107236bb6">
+<files psalm-version="5.4.0@62db5d4f6a7ae0a20f7cc5a4952d730272fc0863">
   <file src="src/CallbackStream.php">
     <ImplementedReturnTypeMismatch occurrences="1">
       <code>null|callable</code>
@@ -114,10 +114,6 @@
     <MoreSpecificImplementedParamType occurrences="1">
       <code>$requestTarget</code>
     </MoreSpecificImplementedParamType>
-    <PossiblyNullOperand occurrences="2">
-      <code>$this-&gt;uri-&gt;getPort()</code>
-      <code>$uri-&gt;getPort()</code>
-    </PossiblyNullOperand>
   </file>
   <file src="src/Response.php">
     <DocblockTypeContradiction occurrences="5">
@@ -169,11 +165,6 @@
     <MixedAssignment occurrences="1">
       <code>$hasContentType</code>
     </MixedAssignment>
-  </file>
-  <file src="src/Response/JsonResponse.php">
-    <UnusedFunctionCall occurrences="1">
-      <code>json_encode</code>
-    </UnusedFunctionCall>
   </file>
   <file src="src/Response/RedirectResponse.php">
     <DocblockTypeContradiction occurrences="1">
@@ -234,21 +225,12 @@
     </RedundantConditionGivenDocblockType>
   </file>
   <file src="src/ServerRequestFilter/FilterUsingXForwardedHeaders.php">
-    <ImpureMethodCall occurrences="7">
+    <ImpureMethodCall occurrences="4">
       <code>getHeaderLine</code>
       <code>getServerParams</code>
       <code>getUri</code>
-      <code>withHost</code>
-      <code>withPort</code>
-      <code>withScheme</code>
       <code>withUri</code>
     </ImpureMethodCall>
-    <LessSpecificReturnStatement occurrences="1">
-      <code>$proxyCIDRList</code>
-    </LessSpecificReturnStatement>
-    <MoreSpecificReturnType occurrences="1">
-      <code>list&lt;non-empty-string&gt;</code>
-    </MoreSpecificReturnType>
   </file>
   <file src="src/Stream.php">
     <PossiblyNullArgument occurrences="4">
--- a/psalm.xml.dist
+++ b/psalm.xml.dist
@@ -5,6 +5,8 @@
     xmlns="https://getpsalm.org/schema/config"
     xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
     errorBaseline="psalm-baseline.xml"
+    findUnusedCode="false"
+    findUnusedPsalmSuppress="true"
 >
     <projectFiles>
         <directory name="src"/>
--- a/psalm.xml.dist
+++ b/psalm.xml.dist
@@ -14,6 +14,10 @@
         </ignoreFiles>
     </projectFiles>
 
+    <stubs>
+        <file name="psalm/http-message-stubs/UriInterface.phpstub"/>
+    </stubs>
+
     <issueHandlers>
         <DeprecatedFunction>
             <errorLevel type="suppress">
--- a/psalm/http-message-stubs/UriInterface.phpstub
+++ b/psalm/http-message-stubs/UriInterface.phpstub
@@ -1,0 +1,18 @@
+<?php
+declare(strict_types=1);
+
+namespace Psr\Http\Message;
+
+/**
+ * [...]
+ * Instances of this interface are considered immutable; all methods that
+ * might change state MUST be implemented such that they retain the internal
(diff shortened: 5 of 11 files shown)


boesing:
TBH: Not even sure if it makes any sense to pass a port via `Host` header. When I connect via TCP to port 443 and add port 80 to the `Host` header, it will still connect to 443 tho...
I wonder if any webserver out there then tries to internally connect to port 80. I hope they won't as this could be a serious security problem when some1 can use a webserver to actually connect to another port internally.

I guess most if not any webserver (besides `apache2`) do not pass any port via `X-Forwarded-Host` header but since there are already 2 reports regarding the same bug, I'd say lets fix this.

src/ServerRequestFilter/FilterUsingXForwardedHeaders.php
```diff
@@ -76,7 +77,8 @@ public function __invoke(ServerRequestInterface $request): ServerRequestInterfac
 
             switch ($headerName) {
                 case self::HEADER_HOST:
-                    $uri = $uri->withHost($header);
+                    [$host] = UriFactory::marshalHostAndPortFromHeader($header);
+                    $uri    = $uri->withHost($host);
```
TimWolla: Interesting, does `withHost()` not reject invalid hostnames containing a `:`?

TimWolla: My first hunch is that this is not a correct change. Instead I would expect that the port is taken from the xfh header if provided, as the reverse proxy in front might only provide xfh, but not xfp.

I'm not quite sure with one should have priority if both are provided, though. Based on my experience helping to maintain the HAProxy issue tracker, I suspect that xfh should have priority, because generally the rule is that the `host` header is the authoritative value for routing. I would want to think about that a little more, though.

(Side note: Reverse proxies that do not pass the `host` header as-is are generally considered broken)

boesing: Unfortunately not.
https://github.com/laminas/laminas-diactoros/blob/b2355dc58253c480aa25ba085845a5aabf84955d/src/Uri.php#L296-L315

boesing: If you ask me, the real problem here is 
> (Side note: Reverse proxies that do not pass the host header as-is are generally considered broken)

Thats definitely one problem, the other is, that some webservers think they have to add the port to the `Host` header even tho, it was not initially sent.

I've tried cURL and my Chrome:

# cURL

```
→ curl --http1.1 -vs https://example.org > /dev/null
*   Trying [2606:2800:220:1:248:1893:25c8:1946]:443...
* Connected to example.org (2606:2800:220:1:248:1893:25c8:1946) port 443 (#0)
* ALPN: offers http/1.1
*  CAfile: /etc/ssl/cert.pem
*  CApath: none
* (304) (OUT), TLS handshake, Client hello (1):
} [313 bytes data]
* (304) (IN), TLS handshake, Server hello (2):
{ [88 bytes data]
* (304) (OUT), TLS handshake, Client hello (1):
} [346 bytes data]
* (304) (IN), TLS handshake, Server hello (2):
{ [155 bytes data]
* (304) (IN), TLS handshake, Unknown (8):
{ [21 bytes data]
* (304) (IN), TLS handshake, Certificate (11):
{ [3106 bytes data]
* (304) (IN), TLS handshake, CERT verify (15):
{ [264 bytes data]
* (304) (IN), TLS handshake, Finished (20):
{ [52 bytes data]
* (304) (OUT), TLS handshake, Finished (20):
} [52 bytes data]
* SSL connection using TLSv1.3 / AEAD-AES256-GCM-SHA384
* ALPN: server accepted http/1.1
* Server certificate:
*  subject: C=US; ST=California; L=Los Angeles; O=Internet?Corporation?for?Assigned?Names?and?Numbers; CN=www.example.org
*  start date: Jan 13 00:00:00 2023 GMT
*  expire date: Feb 13 23:59:59 2024 GMT
*  subjectAltName: host "example.org" matched cert's "example.org"
*  issuer: C=US; O=DigiCert Inc; CN=DigiCert TLS RSA SHA256 2020 CA1
*  SSL certificate verify ok.
> GET / HTTP/1.1
> Host: example.org
> User-Agent: curl/7.86.0
> Accept: */*
```

# Chrome
```
GET / HTTP/1.1
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7
Accept-Encoding: gzip, deflate, br
Accept-Language: de-DE,de;q=0.9,en-US;q=0.8,en;q=0.7
Cache-Control: max-age=0
Connection: keep-alive
DNT: 1
Host: example.org
If-Modified-Since: Thu, 17 Oct 2019 07:18:26 GMT
If-None-Match: "3147526947"
Sec-Fetch-Dest: document
Sec-Fetch-Mode: navigate
Sec-Fetch-Site: none
Sec-Fetch-User: ?1
Upgrade-Insecure-Requests: 1
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/111.0.0.0 Safari/537.36
sec-ch-ua: "Google Chrome";v="111", "Not(A:Brand";v="8", "Chromium";v="111"
sec-ch-ua-mobile: ?0
sec-ch-ua-platform: "macOS"
```

So both actually do not add the port to the `Host` header and thats absolutely correct. The port provided by `Host`-Header has no influence on what is happening via TCP and thus, it is just unnecessary.

But somehow, in [this](https://github.com/laminas/laminas-diactoros/issues/111#issuecomment-1497523633) comment, a user reports that this is happening in NGINX + PHP built-in webserver.

----

I spent some time in creating that scenario using docker `nginx` and installed PHP 8.2 CLI in there.

```
server {
    listen       80;
    server_name  localhost;
    
    location / {
        proxy_pass http://localhost:8081;
    }
}
```

```
$ nginx -S localhost:8081 -t . index.php
```

```php
<?php

var_dump($_SERVER);
```

```
→ curl -v localhost:8080
*   Trying 127.0.0.1:8080...
* Connected to localhost (127.0.0.1) port 8080 (#0)
> GET / HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.86.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Server: nginx/1.23.4
< Date: Wed, 05 Apr 2023 22:05:47 GMT
< Content-Type: text/html; charset=UTF-8
< Transfer-Encoding: chunked
< Connection: keep-alive
< Host: localhost:8081
< X-Powered-By: PHP/8.2.4
< 
array(18) {
  ["DOCUMENT_ROOT"]=>
  string(16) "/usr/share/nginx"
  ["REMOTE_ADDR"]=>
  string(9) "127.0.0.1"
  ["REMOTE_PORT"]=>
  string(5) "55030"
  ["SERVER_SOFTWARE"]=>
  string(28) "PHP 8.2.4 Development Server"
  ["SERVER_PROTOCOL"]=>
  string(8) "HTTP/1.0"
  ["SERVER_NAME"]=>
  string(9) "localhost"
  ["SERVER_PORT"]=>
  string(4) "8081"
  ["REQUEST_URI"]=>
  string(1) "/"
  ["REQUEST_METHOD"]=>
  string(3) "GET"
  ["SCRIPT_NAME"]=>
  string(10) "/index.php"
  ["SCRIPT_FILENAME"]=>
  string(26) "/usr/share/nginx/index.php"
  ["PHP_SELF"]=>
  string(10) "/index.php"
  ["HTTP_HOST"]=>
  string(14) "localhost:8081"
  ["HTTP_CONNECTION"]=>
  string(5) "close"
  ["HTTP_USER_AGENT"]=>
  string(11) "curl/7.86.0"
  ["HTTP_ACCEPT"]=>
  string(3) "*/*"
  ["REQUEST_TIME_FLOAT"]=>
  float(1680732347.931441)
  ["REQUEST_TIME"]=>
  int(1680732347)
}
* Connection #0 to host localhost left intact
```

So actually, NGINX is adding the port to the `Host` header when using `proxy_pass`. But it actually contains the port which was used by NGINX to communicate with the upstream host. So the problem is, when you are forwarding HTTPS => non-HTTPS, that `Host` will contain port 80 but diactoros is meant to detect the actualy Port 443 which was initially used to connect to NGINX and not which was used to the upstream. Even tho, the config is invalid, I'd say that the port from the `Host` header should just get ignored as it most probably contains crap. I can't think of a scenario where this port would be of any interest 🤔 

Once I add `proxy_set_header Host localhost;`, the `Host` header within PHP does not contain the port from `proxy_pass` anymore. So yes, the best way would be to actually pass the initial `Host` via `Host` header, but not sure if that is something which can be configured for **traefik** (which was reported as sending the `Host` header containing the port initially).

boesing: TL;DR: Preserving the port from `X-Forwarded-Host` in case `X-Forwarded-Port` is not available.

Xerkus: @boesing try the same but with non-default port. eg with 8080.

I know that traefik and envoy are appending port to x-forwarded-host when it is not default for the proto (essentially if it appears in uri). Pretty much how `Host` header is defined for HTTP/1.1. They also always send x-forwarded-port.

weierophinney (approved):
This looks... pretty sane. My one concern is how this plays with our fixes for [LP-2022-02](https://getlaminas.org/security/advisory/LP-2022-02}, though what it looks like in here is it is properly considering X-Forwarded-Port as the "source of truth" if both that and X-Forwarded-Host are present.

src/ServerRequestFilter/FilterUsingXForwardedHeaders.php
TimWolla: Thank you. I've checked the nginx source code and indeed it uses `$proxy_host` as the default value for the `host` header of the forwarded request, not `$host`. You generally have to use `proxy_set_header host $host;` to obtain a non-broken nginx configuration.

Leaving a reference to the nginx source code for posterity: https://hg.nginx.org/nginx/file/5f1d05a21287/src/http/modules/ngx_http_proxy_module.c#l823

Will have a look at Traefik later as well.

TimWolla: > The port provided by Host-Header has no influence on what is happening via TCP and thus, it is just unnecessary.

This is not true. The port within the `host` header is absolutely necessary for gateways (“reverse proxies”) that act as a dumb pipe to work correctly, because the TCP-level port information will be lost in the presence of a gateway. For a non-default port, the user-agent (“browser”) will include the port within the `host` header. The optional port as part of `host` header is specified in RFC 9110: https://httpwg.org/specs/rfc9110.html#field.host

TimWolla:
> though what it looks like in here is it is properly considering X-Forwarded-Port as the "source of truth" if both that and X-Forwarded-Host are present.

@weierophinney From my understanding, the priority is implicitly decided based on the ordering of the headers within the `$trustedHeaders` list. I find this reasonable, but it should probably be spelled out explicitly that the order of headers matters (if this is the behavior we want to guarantee here).

-------

The current version of the PR looks reasonable to me as well, the port is taken from the xfh header (if available), if both xfh and xfp match, it does the right thing by definition and if they differ, then the user is able to select the priority based on the ordering in the list. (If xfh and xfp differ, the reverse proxy is broken beyond repair, so … who cares).

Please just give me the time to also check the Traefik source code / behavior (see the existing comment thread with boesing), so that the decision can be backed up with some hard evidence in case someone complains in the future.

src/ServerRequestFilter/FilterUsingXForwardedHeaders.php
TimWolla: Traefik by default passes the host header as-is and *also* adds xfh and xfp. xfp is derived from the host:

https://github.com/traefik/traefik/blob/8174860770e536b4afb541e0ab13b3611a101430/pkg/middlewares/forwardedheaders/forwarded_header.go#L119

Specifically if the incoming `host` header does not contain a port, Traefik will not add one to xfh:

```
GET / HTTP/1.1
Host: example.com
User-Agent: curl/7.68.0
Accept: */*
X-Forwarded-For: 127.0.0.1
X-Forwarded-Host: example.com
X-Forwarded-Port: 80
X-Forwarded-Proto: http
X-Forwarded-Server: *redacted*
X-Real-Ip: 127.0.0.1
Accept-Encoding: gzip
```

The [`passHostHeader`](https://doc.traefik.io/traefik/v2.1/routing/services/#pass-host-header) config can be used to set `host` as the backends hostname (but xfh and xfp will still be correctly derived from the incoming request).

TimWolla (approved):
The proposed behavior looks correct to me. There should likely be a follow-up that makes `->withHost()` more strict to prevent it from accepting garbage (such as colons or whitespace). See also this comment in a previous PR of mine: https://github.com/laminas/laminas-diactoros/pull/97#issuecomment-1172360058

PR #136, opened 2023-04-05
Allow usage with PSR-7 v2

weierophinney:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes
| New Feature   | yes
| RFC           | no
| QA            | yes

### Description

This patch adapts Diactoros to work with either PSR-7 v1.1 or v2.0 by explicitly adding the parameter type hints defined in PSR-7 v2.0, as well as `void` return type declarations where not previously declared.

This patch depends on the following upstream changes:

- php-fig/http-factory#13 (will likely be tagged 1.1)
- php-http/psr7-integration-tests#68 (will likely be tagged 1.3)

This patch should be applied to an upcoming 3.0.x branch, for release as 3.0, due to the following BC breaks:

- Addition of explicit parameter type declarations (where known) to all PSR-7 implementation signatures. This change will break any extensions of these classes, as the signatures have changed in a way that does not conform to covariance rules for return types. Specifically, any method that was updated to return `void` in the PSR-7 spec that we had not marked in our own implementation with a `void` return declaration leads to the BC break, as updating our implementation means extensions that do not declare it will now cause a fatal error.

### Tasks

The following are required before removing this from draft status:

- [x] Merge and release of [PSR-7 integration test suite changes](https://github.com/php-http/psr7-integration-tests/pull/68)
- [x] Merge and release of [PSR-17 updates](https://github.com/php-fig/http-factory/pull/13)
- [x] Removal of temporary repositories used to test against the above two dependencies


composer.json
```diff
@@ -33,10 +33,20 @@
...
+    "repositories": {
+        "http-factory": {
+            "type": "vcs",
+            "url": "https://github.com/ghostwriter/http-factory.git"
+        },
+        "integration-tests": {
+            "type": "vcs",
+            "url": "https://github.com/weierophinney/psr7-integration-tests"
+        }
+    },
```
Ocramius: To be dropped before merge

composer.json
```diff
@@ -33,10 +33,20 @@
...
     "require": {
         "php": "~8.0.0 || ~8.1.0 || ~8.2.0",
-        "psr/http-factory": "^1.0",
-        "psr/http-message": "^1.0"
+        "psr/http-factory": "dev-feature/http-message-v2 as 1.1.0",
```
Ocramius: To be adjusted before merge

composer.json
```diff
@@ -45,7 +55,7 @@
         "ext-libxml": "*",
         "http-interop/http-factory-tests": "^0.9.0",
         "laminas/laminas-coding-standard": "^2.4.0",
-        "php-http/psr7-integration-tests": "^1.2",
+        "php-http/psr7-integration-tests": "dev-feature/psr-7-type-hints as 1.3.0",
```
Ocramius: To be adjusted before merge

psalm.xml.dist
```diff
@@ -4,6 +4,8 @@
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xmlns="https://getpsalm.org/schema/config"
     xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
+    findUnusedBaselineEntry="true"
+    findUnusedCode="false"
```
Ocramius: We should most likely use `true` here, even if it expands the baseline

src/CallbackStream.php
```diff
@@ -126,7 +122,7 @@ public function isWritable(): bool
     /**
      * {@inheritdoc}
      */
-    public function write($string): void
+    public function write(string $string): int
```
Ocramius, line 125: Will `v1` still work, with these types applied? :thinking:

composer.json
weierophinney: Yep - that's why this is in draft state. :smile:

src/CallbackStream.php
weierophinney: Yes: https://github.com/laminas/laminas-diactoros/actions/runs/4629754113/jobs/8190430823?pr=136#step:3:387 (that demonstrates installation of 1.1 for a successful PHPUnit run).

psalm.xml.dist
weierophinney: I'll update to do that.

weierophinney: Updated. :)

composer.json
Ocramius: Figured: just making sure it isn't forgotten :+1:

TimWolla:
> This change will break any extensions of these classes, as the signatures have changed in a way that does not conform to contravariance rules for parameters.

I do not believe this description is accurate. A deriving class may leave out parameter types of a parent class (“widen the type”), because everything that may be passed to the parent class may still be passed to the deriving class: https://3v4l.org/gfWBN

However adding parameter types may break consumers that rely on passing an invalid type. However to my understanding, Diactoros already included checks to validate the type.

So, I believe this is safe for a minor, unless I'm missing something?

src/CallbackStream.php
```diff
@@ -160,7 +156,7 @@ public function getContents(): string
     /**
      * {@inheritdoc}
      */
-    public function getMetadata($key = null)
+    public function getMetadata(?string $key = null)
```
ADmad, line 159: Why isn't `mixed` return type used here (and other functions which returned mixed) ?

ADmad: I see that it's because `mixed` type is not added for the interface method `StreamInterface::getMetaData()`. Then my question would be why wasn't it added there?

ADmad: To answer myself it's cause `php-fig/http-message` 2.0 is still PHP 7.2+ :man_facepalming:. https://github.com/php-fig/http-message/blob/2.0/composer.json#L14

weierophinney:
@TimWolla —

> So, I believe this is safe for a minor, unless I'm missing something?

You were right about the contravariance versus covariance bit; I had that part wrong in the description and have since corrected it.

The problem is actually with the return type hints: when we were defining return type hints in this implementation, we missed a few cases that later became `void` in the PSR-7 v2 spec. Going from implied `mixed` to `void` is the BC break. If an extension of a method now defined to return `void` returns something or omits the return type declaration, PHP will raise a fatal error.

psalm.xml.dist.orig
Xerkus, line 1: Accidentally committed backup

boesing:
I wonder if we should pin php-http integration tests to specific versions anyway. Since they can change upstream, it might lead to failing checks in PRs which are unrelated to those changes. Having renovate bumping these commits could help, no?

psalm.xml.dist.orig
weierophinney: Fixed.

Xerkus:
> Having renovate bumping these commits could help, no?

That would only affect latest branch.
Integration tests could be updated to cover newly discovered cross-compatibility issue and we would want CI fail for older branches too, imo.

boesing:
> That would only affect latest branch.

More than enough for PRs.

> Integration tests could be updated to cover newly discovered cross-compatibility issue and we would want CI fail for older branches too, imo.

Having scheduled build for that where we explictly use `composer require integration-tests:dev-master` should be used for discovering these kinds of issues.

Xerkus:
Both dependencies got releases. `1.0.2` and `1.3.0` respectively

weierophinney:
I've now created the 3.0.x branch, and updated this patch to be against that branch. Additionally, both psr/http-factory and php-http/psr7-integration-tests have released stable versions that support PSR-7 1.1 and 2.0, allowing me to rebase my patch to reference the stable branches.

Ready for final review and release! Once done, I'll turn my efforts to laminas/laminas-stratigility, and then the various Mezzio components.

test/MessageTraitTest.php
```diff
@@ -35,35 +35,6 @@ public function testProtocolMutatorReturnsCloneWithChanges(): void
...
-            'null'                 => [null],
-            'true'                 => [true],
-            'false'                => [false],
-            'int'                  => [1],
-            'float'                => [1.1],
-            'array'                => [['1.1']],
-            'stdClass'             => [(object) ['version' => '1.0']],
-            '1-without-minor'      => ['1'],
-            '1-with-invalid-minor' => ['1.2'],
-            '1-with-hotfix'        => ['1.2.3'],
```
TimWolla: Should these remain?

test/ResponseTest.php
```diff
@@ -152,12 +152,18 @@ public function ianaCodesReasonPhrasesProvider(): array
             $value       = $value;
             $description = $description;
```
Xerkus: What this assignment achieves?

test/UriTest.php
```diff
@@ -166,8 +155,8 @@ public function testWithPortReturnsNewInstanceWithProvidedPort($port): void
     public function testWithPortReturnsSameInstanceWithProvidedPortIsSameAsBefore(): void
     {
         $uri = new Uri('https://user:pass@local.example.com:3001/foo?bar=baz#quz');
-        /** @psalm-suppress InvalidArgument */
-        $new = $uri->withPort('3001');
+        /** @psalm-suppress PossiblyInvalidArgument,InvalidArgument */
```
Xerkus: Unused suppression

test/ResponseTest.php
```diff
@@ -166,10 +172,9 @@ public function ianaCodesReasonPhrasesProvider(): array
 
     /**
      * @dataProvider ianaCodesReasonPhrasesProvider
-     * @param numeric-string $code
      * @param non-empty-string $reasonPhrase
      */
-    public function testReasonPhraseDefaultsAgainstIana(string $code, string $reasonPhrase): void
+    public function testReasonPhraseDefaultsAgainstIana(int $code, string $reasonPhrase): void
     {
         /** @psalm-suppress InvalidArgument */
```
Xerkus: Unused suppression

test/ResponseTest.php
```diff
@@ -231,27 +211,24 @@ public function testConstructorCanAcceptAllMessageParts(): void
 
     /**
      * @dataProvider validStatusCodes
-     * @param int|numeric-string $code
      */
-    public function testCreateWithValidStatusCodes($code): void
+    public function testCreateWithValidStatusCodes(int $code): void
     {
         /** @psalm-suppress PossiblyInvalidArgument */
```
Xerkus: Unused suppression

boesing (approved):
LGTM

test/MessageTraitTest.php
weierophinney: That's a good catch; updating now!

test/ResponseTest.php
weierophinney: Don't know. It's not part of the changes I introduced, but I'll remove them.

test/UriTest.php
weierophinney: Resolved

test/ResponseTest.php
weierophinney: Resolved

test/ResponseTest.php
weierophinney: Resolved

Ocramius (approved):
🚢 

Thanks @weierophinney!

Xerkus (changes requested):
https://github.com/laminas/laminas-diactoros/blob/22eca5bd22b437e129cc1e1392e888bb07bdedda/src/RequestTrait.php#L156
Missing parent type declaration.

---
Some tangentially related comments:

https://github.com/laminas/laminas-diactoros/blob/22eca5bd22b437e129cc1e1392e888bb07bdedda/src/UploadedFile.php#L60-L62
Interface defines size as nullable. This should probably be changed to nullable too. If so this will need to be backported to 2.x

https://github.com/laminas/laminas-diactoros/blob/22eca5bd22b437e129cc1e1392e888bb07bdedda/src/MessageTrait.php#L353
`MessageTrait::validateProtocolVersion()` could have argument type defined and is_string check dropped to match `withProtocolVersion()`.

https://github.com/laminas/laminas-diactoros/blob/22eca5bd22b437e129cc1e1392e888bb07bdedda/src/ServerRequest.php#L176

This uses parameter name `$attribute` and interface uses `$name`. Should be changed for consistency. Same for `withAttribute()`

https://github.com/laminas/laminas-diactoros/blob/22eca5bd22b437e129cc1e1392e888bb07bdedda/src/StreamFactory.php#L31
Interface uses `$filename`

weierophinney:
> Missing parent type declaration.

Thanks - updating that now.

weierophinney:
> Interface defines size as nullable. This should probably be changed to nullable too. If so this will need to be backported to 2.x

While the size _can_ be nullable, we're saying in this particular _implementation_ that it must be an integer. Since this narrows what we return from `getSize()`, it's still valid and fulfills the spec. I see no reason to change this currently.

Xerkus:
> While the size _can_ be nullable, we're saying in this particular _implementation_ that it must be an integer.

This somewhat conflicts with the argument of UriFactory as using null will lead to type error
https://github.com/laminas/laminas-diactoros/blob/22eca5bd22b437e129cc1e1392e888bb07bdedda/src/UploadedFileFactory.php#L18-L30

weierophinney:
> https://github.com/laminas/laminas-diactoros/blob/22eca5bd22b437e129cc1e1392e888bb07bdedda/src/ServerRequest.php#L176
> 
> This uses parameter name `$attribute` and interface uses `$name`. Should be changed for consistency. Same for `withAttribute()`
> 
> https://github.com/laminas/laminas-diactoros/blob/22eca5bd22b437e129cc1e1392e888bb07bdedda/src/StreamFactory.php#L31
> 
> 
> Interface uses `$filename`

I think these should be done with the 2.x series, and then merged up.

weierophinney:
> > While the size _can_ be nullable, we're saying in this particular _implementation_ that it must be an integer.
> 
> This somewhat conflicts with the argument of UriFactory as using null will lead to type error

Alright, that argument I can get behind. I'll make it nullable, but not have a default value.

Xerkus:
> Alright, that argument I can get behind. I'll make it nullable, but not have a default value.

Will you make PR against 2.x or should I?

Xerkus (approved):

weierophinney:
@Xerkus —

I've pushed changes for each of the following issues you flagged:

- Added the missing `string` parameter type declaration to `RequestTrait::withRequestTarget()`
- Added a `string` parameter type declaration to `MessageTrait::validateProtocolVersion()`'s `$version` argument, and removed the unneeded conditional that checked for a string in that method.
- Made the `$size` argument/property in `UploadedFileFactoryInterface` nullable.

I've been thinking about the parameter name changes, and it DOES NOT make sense to do them on the 2.x branch, as, with named arguments in PHP 8, this would be a BC break (name changes). While it could be considered a bugfix, it would potentially break code for users. So let's do it here; I'll push those shortly.

Xerkus:
@weierophinney thank you!

PR #141, opened 2023-04-13
Disallow cloning Stream objects

TimWolla:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | ?
| BC Break      | ?
| New Feature   | no
| RFC           | yes
| QA            | no

### Description

Not necessarily intended to be merged, primarily intended to put this up for discussion with an actual code change instead of just raising an issue. I did not encounter this in a real-world app, this was more of a “what if”.

------------

Streams hold a reference to the stateful resource handle for their actual contents. Cloning a Stream will not actually clone the underlying resource, thus both streams would still refer to the same resource after cloning and any changes in one stream object would be reflected in the other object. This violates user expectations after a cloning operation.

Disallow cloning entirely as the safe default choice. Alternatively a new stream could be created and attached and the contents could be copied over. This can get expensive with larger or infinite streams, though.

### DIFF
--- a/test/StreamTest.php
+++ b/test/StreamTest.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace LaminasTest\Diactoros;
 
 use CurlHandle;
+use Error;
 use GdImage;
 use InvalidArgumentException;
 use Laminas\Diactoros\Stream;
@@ -678,5 +679,19 @@ final class StreamTest extends TestCase
         $resource = fopen('php://input', 'r');
         $stream   = new Stream($resource);
         $this->assertNull($stream->getSize());
+    }
+
+    public function testStreamsAreUnclonable(): void
+    {
+        $stream = new Stream(fopen('php://temp', 'r+'));
+        $stream->write('foo');
+
+        $this->assertSame('foo', $stream->__toString());
+
+        $this->expectException(Error::class);
+        $this->expectExceptionMessage('private Laminas\Diactoros\Stream::__clone()');
+
+        /** @psalm-suppress InvalidClone */
+        clone $stream;
     }
 }
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -364,4 +364,11 @@ class Stream implements StreamInterface, Stringable
 
         return false;
     }
+
+    /**
+     * Disallow stream cloning.
+     */
+    private function __clone()
+    {
+    }
 }


froschdesign:
@TimWolla 
> Not necessarily intended to be merged, primarily intended to put this up for discussion…

Then I convert this pull request into a [draft](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests#draft-pull-requests).

Ocramius:
The proposal is very much sensible: due to streams being a squishy mess for the foreseeable future, it's best to not allow multiple references to the same stream in different objects, and cloning should be disallowed.

The newly introduced test could perhaps even be dropped :thinking:

Ocramius (approved):
The proposal is very much sensible: due to streams being a squishy mess for the foreseeable future, it's best to not allow multiple references to the same stream in different objects, and cloning should be disallowed.

The newly introduced test could perhaps even be dropped :thinking:

test/StreamTest.php
```diff
@@ -679,4 +680,17 @@ public function testSizeReportsNullForPhpInputStreams(): void
...
+    {
+        $stream = new Stream(fopen('php://temp', 'r+'));
+        $stream->write('foo');
+
+        $this->assertSame('foo', $stream->__toString());
+
+        $this->expectException(Error::class);
+        $this->expectExceptionMessage('private Laminas\Diactoros\Stream::__clone()');
+
+        clone $stream;
```
gsteel:
```suggestion

        /** @psalm-suppress InvalidClone */
        clone $stream;
```

gsteel (approved):
LGTM! This would be considered a BC break no?

TimWolla:
> This would be considered a BC break no?

This technically is a compatibility break, yes. That's the primary reason why I've mentioned the “put up for discussion” part. I'll leave it to you to decide whether this is a bugfix or a debugging aid and whether the compatibility is relevant or not.

Ocramius:
Yes, this is a potential BC break, but cloning a stream is really asking for trouble.

Xerkus:
This looks like something that needs to be brought to FIG to explicitly define expected clone behavior.

I see a benefit for well defined clone behavior where stream is expected to do internal `stream_copy_to_stream()`. May be for original stream object too if resource is not rewindable.

This also raises question about cloning request/response objects. Stream is mutable and it will be shared between the two.

Ocramius:
Very good point made by @Xerkus: the removal of `clone` support on our `Stream` could easily break other message implementations :thinking: 

Is it possible to implement safe cloning, until we disallow it in FIG itself? :thinking:

TimWolla:
> Is it possible to implement safe cloning, until we disallow it in FIG itself? thinking

Depends: Infinite streams are unsafe, but one could tell users “don't do that”. I'm also not sure about unrewindable generator-based streams.

Anyway: If your concerned about interoperability then even “safe cloning” is unsafe, because an implementation might actually expect the underlying stream resource to remain the same after a clone. This is an … odd assumption, but still an assumption that might've been made.

Xerkus:
> This is an … odd assumption, but still an assumption that might've been made.

This is why this needs to be addressed at FIG level since whatever approach is taken it will defy one expectation or another.

PR #146, opened 2023-05-01
Drop legacy helper functions in Zend namespace

Xerkus:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes
| New Feature   | no
| RFC           | no
| QA            |no

### Description

Legacy helper functions in Zend namespace were left in place to facilitate migration but no longer needed.


Ocramius (approved):

Xerkus:
Psalm issue: `UndefinedAttributeClass` can be suppressed with issue handler using targeted `referencedClass`.

When running on 8.2 psalm targets 8.0 and complains with

```
ERROR: InvalidAttribute - src/Uri.php:246:11 - The class SensitiveParameter doesn't have the Attribute attribute (see https://psalm.dev/242)
        #[SensitiveParameter]
```
This issue does not have `referencedClass` to make targeted issue handler and it can't be baselined either.

psalm.xml.dist
```diff
@@ -50,6 +50,11 @@
                 <referencedClass name="GdImage" />
             </errorLevel>
         </UndefinedClass>
+        <UndefinedAttributeClass>
+            <errorLevel type="suppress">
```
boesing: should be added to baseline. reason is, that explicitly excluded issues are not (yet) reported by psalm once their unnecessary. 

Not a hard requirement but thats why I tend to put these to baseline

Xerkus: Not baselined because result differs with php version psalm runs on.

Running on php 8.2 with --update-baseline removes baseline entry as unused which is how it got caught by CI here.
It still fails locally with error mentioned above but baselining that would be reported as unused in CI since it runs psalm on php 8.0.

boesing: psalm is only executed on lowest locked deps, no clue why u have different errors at all unless dropping support for php versions or bumping psalm deps.

Xerkus: Because locally psalm runs on local php version

Ocramius: TBH, the baseline is PHP version specific: I generate it with `docker run --rm -ti -v $(pwd):/project -w /project php:$VERSION` or such.

Ocramius: To rephrase: agree with @boesing that this should be in baseline

Xerkus: This **can't be in baseline** because as soon as contributor runs `composer static-analysis -- --update-baseline` on php 8.2 it will be dropped from baseline, because when psalm runs on 8.2 the reported issue is different.

We can't really expect contributors to use specific php version needed by CI. This was the issue with lock file and why `composer.platform.php` was introduced.

Ocramius: It **must** be in baseline, because:

 1. you run that command on the precise version you are testing it on (not your local PHP version)
 2. it should be disappearing the moment we drop PHP 8.1 support in the future

This applies to all general errors that depend on the PHP version, such as when PHP improves the types on a specific function.

psalm-baseline.xml
```diff
@@ -269,9 +269,6 @@
     <PossiblyNullOperand>
       <code><![CDATA[$this->port]]></code>
     </PossiblyNullOperand>
-    <UndefinedAttributeClass>
-      <code>SensitiveParameter</code>
-    </UndefinedAttributeClass>
```
Ocramius: To be reverted here, otherwise all good :+1:

Ocramius (approved):

psalm-baseline.xml
Xerkus: Oof. I did the opposite on rebase:  reverted other changes and kept this one

Ocramius:
Thanks @Xerkus!

Issue #57, opened 2020-11-17
Stream is not fully compatible with GD resource


boesing:
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 2.4.0

#### Summary

Using a GD resource works fine until there is interaction with the stream.

#### Current behavior

When working with GD Images, most of the methods of the Stream (with PHP 8.0 its `GDImage`) will result in errors.

#### How to reproduce

```
    public function testIsReableWithGdResource()
    {
        $resource = imagecreate(1, 1);
        $stream   = new Stream($resource);
        $this->assertTrue($stream->isReadable());
    }

    public function testCloseWithGdResource()
    {
        $resource = imagecreate(1, 1);
        $stream   = new Stream($resource);
        $this->assertNull($stream->close());
    }

```
result in 
```
1) LaminasTest\Diactoros\StreamTest::testIsReableWithGdResource
stream_get_meta_data(): supplied resource is not a valid stream resource

2) LaminasTest\Diactoros\StreamTest::testCloseWithGdResource
fclose(): supplied resource is not a valid stream resource
```

So the existing code is already broken and doesn't work with image resources.

#### Expected behavior

Using methods return proper informations such as the GD Image/Resource is not writable, seekable, e.g.

_Originally posted by @ADmad in https://github.com/laminas/laminas-diactoros/pull/46#r516719793_

settermjd:
I'm happy to work on this

settermjd:
I don't know that it's possible to fix this, because stream_get_meta_data only works with streams created by fopen(), fsockopen(), and pfsockopen(). I'm open to learning about workarounds, but don't know of any atm.

ADmad:
I don't think the issue is fixable either. 

Two options I see are:
1) Deprecate and drop support for using GD resource / `GDImage` instance with `Stream`. 
2) Throw an exception when trying to do operations like seeking which are not possible with a `GDImage` instance.

settermjd:
Might be the case. I have to add some feedback from @weierophinney, tomorrow.

settermjd:
Just found it:

[...]

PR #147, opened 2023-05-02
ImageStream implementation

weierophinney:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | future
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

Per discussion in #57, this patch does the following:

- Introduces the `ImageStream` class as an extension of `Stream` that only works with `GdImage` resources.
- Updates `Stream` to emit an `E_USER_DEPRECATED` error when used with a `GdImage` resource, pointing users to the new `ImageStream` implementation.
- Updates `StreamFactory::createStreamFromResource()` to detect `GdImage` resources, and to create and return an `ImageStream` when it does.

The plan will be for 3.0.0 to remove support for `GdImage` from `Stream`.

### DIFF
--- a/src/ImageStream.php
+++ b/src/ImageStream.php
@@ -1,0 +1,63 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laminas\Diactoros;
+
+use GdImage;
+
+use function sprintf;
+
+/**
+ * This class purposely does not override the $resource property in order to
+ * allow it to extend Stream. If defined here with a type, PHP will raise a
+ * fatal error complaining that it must not define a type for the property.
+ */
+class ImageStream extends Stream
+{
+    /**
+     * This purposely does not explicitly override the $resource property in
+     * order to allow it to extend Stream. If defined here with a type, PHP will
+     * raise a fatal error complaining that it must not define a type for the
+     * property.
+     *
+     * @var null|GdImage
+     */
+    protected $resource;
+
+    public function __construct(
+        GdImage $resource,
+    ) {
+        $this->resource = $resource;
+    }
+
+    /**
+     * {@inheritdoc}
+     * @return null|GdImage
+     */
+    public function detach(): ?GdImage
+    {
+        $resource       = $this->resource;
+        $this->resource = null;
+        return $resource;
+    }
+
+    /**
+     * Attach a new stream/resource to the instance.
+     *
+     * @param GdImage $resource
+     * @param string $mode Unused
+     * @throws Exception\InvalidArgumentException When provided a non-GdImage resource.
+     */
+    public function attach($resource, string $mode = 'r'): void
+    {
+        if (! $resource instanceof GdImage) {
+            throw new Exception\InvalidArgumentException(sprintf(
+                'When attaching a resource to %s, resource must be a GdImage',
+                $this::class,
+            ));
+        }
+
+        $this->resource = $resource;
+    }
+}
--- a/src/ImageStream.php
+++ b/src/ImageStream.php
@@ -33,6 +33,7 @@ class ImageStream extends Stream
 
     /**
      * {@inheritdoc}
+     *
      * @return null|GdImage
      */
     public function detach(): ?GdImage
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -28,7 +28,9 @@ use function sprintf;
 use function stream_get_contents;
 use function stream_get_meta_data;
 use function strstr;
+use function trigger_error;
 
+use const E_USER_DEPRECATED;
 use const SEEK_SET;
 
 /**
@@ -359,6 +361,14 @@ class Stream implements StreamInterface, Stringable
         }
 
         if ($resource instanceof GdImage) {
+            trigger_error(
+                sprintf(
+                    'When using GdImage resources, use %s; %s will drop support for GdImage in 3.0.0',
+                    ImageStream::class,
+                    self::class
+                ),
+                E_USER_DEPRECATED,
+            );
             return true;
         }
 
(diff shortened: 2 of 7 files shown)


src/ImageStream.php
```diff
@@ -0,0 +1,64 @@
...
+use GdImage;
+
+use function sprintf;
+
+/**
+ * This class purposely does not override the $resource property in order to
+ * allow it to extend Stream. If defined here with a type, PHP will raise a
+ * fatal error complaining that it must not define a type for the property.
+ */
+class ImageStream extends Stream
```
Xerkus, line 16: It should not extend Stream at all. It is not compatible with any of its methods.

src/Stream.php
```diff
@@ -359,6 +361,14 @@ private function isValidStreamResourceType(mixed $resource): bool
         }
 
         if ($resource instanceof GdImage) {
```
Xerkus, line 363: I get type errors for pretty much all methods in Stream when GdImage is used as resource.

```
Uncaught TypeError: stream_get_contents(): Argument #1 ($stream) must be of type resource, GdImage given
Uncaught TypeError: stream_get_meta_data(): Argument #1 ($stream) must be of type resource, GdImage given
Uncaught TypeError: fseek(): Argument #1 ($stream) must be of type resource, GdImage given
```
and so on

src/StreamFactory.php
```diff
@@ -35,9 +36,15 @@ public function createStreamFromFile(string $file, string $mode = 'r'): StreamIn
 
     /**
      * {@inheritDoc}
+     *
+     * @param resource|GdImage $resource
      */
```
Xerkus, line 41: I am not sure if this is good. Expanding parameter types is considered type safe. The problem is anyone needing `GdImage` support can no longer depend on PSR factory interface because that would not be type safe.

Wouldn't that go against interoperability goal between libraries since this will be creating dependency on diactoros and this implementation?

weierophinney: `GdImage` is considered an _opaque resource type_. The general assumption of PSR-7 and PSR-17's stream support is that it can work with _resources_ of any type. However, _both_ pre-date the opaque resource types.

This suggests two things to me:

- They should allow usage with opaque resource types _where they make sense_. Since you can have image resources spit out content, I'd argue they make sense here, and, in some ways, are better than normal file-based resources where you lose idempotency.
- Alternately, PSR-7 and/or PSR-17 need updates to either explicitly allow or disallow usage with opaque resource types.

For now, I don't think this patch violates interoperability, as PSR-7 and PSR-17 _do_ talk about resources, and this is a type of resource.

src/Stream.php
Xerkus: It did not work at all for a long time. I suggest "support" is dropped without deprecation as no actual users will be affected.

src/ImageStream.php
boesing: I agree here, extending `Stream` makes no sense as we have to handle almost every method in a specific way.
Especially due to the `GdImage`/`resource` problematic, we could also just provide `GdImageStream` as it is for `GdImage` only. 
Not sure if some1 wants to use this for anything else, so renaming it might be helpful but I do not insist on this.

test/StreamTest.php
```diff
@@ -67,12 +71,29 @@ public function testCanInstantiateWithStreamResource(): void
         $this->assertInstanceOf(Stream::class, $stream);
     }
 
+    /**
+     * @todo For version 3, this test should verify that a GdImage raises an  exception.
```
boesing, line 75: with v3 we should verify if either `string` or `resource` is passed, **everything** else should throw an exception.

src/Stream.php
boesing: Yah, there were no proper tests for PHP 8.1
Thats why I raised the whole concerns in the first place.

Also stream seeking on `gdimage` resources never worked and plenty of other stuff (i.e. writing to that stream).

Thats why the new `ImageStream` implementation has to return `false` for `isSeekable`, `isWritable`, etc.

Has to throw exceptions for `getMetadata`, etc.

src/StreamFactory.php
boesing: I would not support `GdImage` via `StreamFactory` TBH

PSR is clear here:
https://github.com/php-fig/http-factory/blob/e616d01114759c4c489f93b099585439f795fe35/src/StreamFactoryInterface.php#L44

If some1 wants to use `GdImage` with `StreamFactoryInterface` they are already lost. If they hardly depend on laminas as an implementation, they don't need PSR and thus can directly instantiate `ImageStream` rather than using the interchangeable factory.  The whole point of using the `StreamFactory` is to provide interchangeability and if we now decide to support `GdImage` that would only fork for laminas but not for guzzle for example. Having these kind of incompatibility is super annoying and thus, I'd say we should not allow that at all.

boesing (changes requested):

boesing:
Summing up Slack discussion:

Currently, neither PHP 7.4 nor 8.0 with both `resource` or `GdImage` are properly functioning as both `GdImage` and the `resource` which are passed do not contain the image type. So should the image created as an PNG or a JPEG?
Lacking this information, using `stream_get_contents` on the `gdimage` resource will result in:

`Warning: stream_get_contents(): supplied resource is not a valid stream resource`

For PHP 8.0 and `GdImage`, that error will slightly change to `Warning: Uncaught TypeError: stream_get_contents(): Argument #1 ($stream) must be of type resource, GdImage given`.

So the whole `image` feature is non-existent and therefore never worked. Won't be even a BC break by removing this feature tho.

weierophinney:
Based on the Slack conversation and the feedback here, I'm going to close this patch, and then open a new one against 3.0.x to drop support for GD images in `Stream`.

PR #148, opened 2023-05-02
Remove support for GD images in Stream implementation

### LINKED ISSUE #57: Stream is not fully compatible with GD resource

### LINKED ISSUE #134: Remove image stream compatibility from `Stream`
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | no
| RFC         | no
| BC Break    | yes

#### Summary

As outlined in #57, the whole image stream functionality should be dropped from the current `Stream` implementation.
We could instead provide an `ImageStream` instead for this as suggested by @weierophinney in https://github.com/laminas/laminas-diactoros/issues/57#issuecomment-1198520464

weierophinney:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | yes
| New Feature   | no
| RFC           | no
| QA            | no

### Description

GdImage-backed `Stream` instances have evidently never worked, but feature removal needs to happen in a new major.

See #57, #134, and #147 for full details behind removal.

### DIFF
--- a/docs/book/v2/api.md
+++ b/docs/book/v2/api.md
@@ -194,7 +194,7 @@ and provides a number of facilities around manipulating the composed PHP stream 
 constructor accepts a stream, which may be one of:
 
 - A stream identifier; e.g., `php://input`, a filename, etc.
-- A PHP stream resource; or
+- A PHP stream resource
 
 If a stream identifier is provided, an optional second parameter may be provided, the file mode by
 which to `fopen` the stream.
--- a/docs/book/v2/api.md
+++ b/docs/book/v2/api.md
@@ -195,7 +195,6 @@ constructor accepts a stream, which may be one of:
 
 - A stream identifier; e.g., `php://input`, a filename, etc.
 - A PHP stream resource; or
-- A GD resource
 
 If a stream identifier is provided, an optional second parameter may be provided, the file mode by
 which to `fopen` the stream.
--- a/docs/book/v3/api.md
+++ b/docs/book/v3/api.md
@@ -1,0 +1,214 @@
+# API
+
+## Request Message
+
+`Laminas\Diactoros\Request` implements [`Psr\Http\Message\RequestInterface`](https://github.com/php-fig/http-message/blob/master/src/RequestInterface.php),
+and is intended for client-side requests. It includes the following methods:
+
+```php
+class Request
+{
+    public function __construct(
+        $uri = null,
+        $method = null,
+        $body = 'php://memory',
+        array $headers = []
+    );
+
+    // See psr/http-message's RequestInterface for other methods
+}
+```
+
+Requests are immutable. Any methods that would change state &mdash; those prefixed with `with` and
+`without` &mdash; all return a new instance with the changes requested.
+
+## ServerRequest Message
+
+For server-side applications, `Laminas\Diactoros\ServerRequest` implements
+[`Psr\Http\Message\ServerRequestInterface`](https://github.com/php-fig/http-message/blob/master/src/ServerRequestInterface.php),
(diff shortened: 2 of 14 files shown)


boesing (changes requested):
Should we re-introduce the stream compat check in `StreamFactory` as it was removed in https://github.com/laminas/laminas-diactoros/pull/45/files#diff-565876e19198a94175bb659207c750dcb80cbbfa104490c6fad5322dd05d5febL49-L53 or do we verify that by waiting for the check in `__construct` => `setStream` of `Stream#__construct`?

weierophinney:
> Should we re-introduce the stream compat check in `StreamFactory` as it was removed in https://github.com/laminas/laminas-diactoros/pull/45/files#diff-565876e19198a94175bb659207c750dcb80cbbfa104490c6fad5322dd05d5febL49-L53 or do we verify that by waiting for the check in `__construct` => `setStream` of `Stream#__construct`?

It's stil there, just done in a different way. In `isValidStreamResourceType`, we check to see if (a) we have a resource, and (b) if it's in the list defined by the class constant `ALLOWED_STREAM_RESOURCE_TYPES`... which now only defines `stream` as a valid resource type.

Keeping it this way allows us to more easily expand it in the future, and keeps the same check that we had prior to adding GD support anyways.

boesing:
> It's stil there, just done in a different way.

`StreamFactory` validates valid resource type by counting on `Stream#__construct` to validate `resource`.
So yes, it is there, but its more implicit than explicit.

That was what I wanted to clarify. If we are happy with implicit validation, thats okay for me as well.

boesing (approved):

Xerkus:
I would like mentions of GD removed from documentation in v2 folder too so it won't be on live site anymore when this is released since it never worked.

Xerkus (approved):

weierophinney:
> I would like mentions of GD removed from documentation in v2 folder too so it won't be on live site anymore when this is released since it never worked.

Done.

docs/book/v2/api.md
```diff
@@ -195,7 +195,6 @@ constructor accepts a stream, which may be one of:
 
 - A stream identifier; e.g., `php://input`, a filename, etc.
 - A PHP stream resource; or
```
Xerkus: One minor change
```suggestion
- A PHP stream resource
```

Xerkus:
Thank you!

PR #153, opened 2023-05-02
Allow numeric header names

### LINKED ISSUE #11: Handle request headers with numeric keys
Here, at getpocket.com, we have had a client hit our servers with header keys as integers. In doing so `HeaderSecurity::assertValidName` is throwing an exception because `! is_string(-1) === true`.

I have been unable to identify any documentation which would suggest that these values could be valid. At this point I believe the best behavior is to ignore these keys.

---

Originally posted by @jeshuaborges at https://github.com/zendframework/zend-diactoros/pull/286

weierophinney:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | yes
| New Feature   | no
| RFC           | no
| QA            | no

### Description

RFC 7230 allows for numeric header names, both integers and floats, though the expectation is that they are provided as string values. Since PHP casts integer strings into integers for purposes of array keys, this can lead to a scenario where `HeaderSecurity` was then flagging the value as invalid - despite the fact that the regex used on strings clearly allows the value. This patch modifies `HeaderSecurity::assertValidName()` to allow for numeric names, and casts them to a string when performing validations.

Marking as a BC break as there is a subtle behavior change.

Fixes #11

### DIFF
--- a/src/AbstractSerializer.php
+++ b/src/AbstractSerializer.php
@@ -8,6 +8,7 @@ use Psr\Http\Message\StreamInterface;
 
 use function array_pop;
 use function implode;
+use function is_int;
 use function preg_match;
 use function sprintf;
 use function str_replace;
@@ -137,11 +138,12 @@ abstract class AbstractSerializer
     /**
      * Filter a header name to wordcase
      *
-     * @param string $header
+     * @param string|int $header
      */
     protected static function filterHeader($header): string
     {
-        $filtered = str_replace('-', ' ', $header);
+        $filtered = is_int($header) ? (string) $header : $header;
+        $filtered = str_replace('-', ' ', $filtered);
         $filtered = ucwords($filtered);
         return str_replace(' ', '-', $filtered);
     }
--- a/src/HeaderSecurity.php
+++ b/src/HeaderSecurity.php
@@ -148,16 +148,16 @@ final class HeaderSecurity
      */
     public static function assertValidName(mixed $name): void
     {
-        if (! is_string($name)) {
+        if (! is_string($name) && ! is_numeric($name)) {
             throw new Exception\InvalidArgumentException(sprintf(
-                'Invalid header name type; expected string; received %s',
+                'Invalid header name type; expected string or numeric value; received %s',
                 is_object($name) ? $name::class : gettype($name)
             ));
         }
-        if (! preg_match('/^[a-zA-Z0-9\'`#$%&*+.^_|~!-]+$/D', $name)) {
+        if (! preg_match('/^[a-zA-Z0-9\'`#$%&*+.^_|~!-]+$/D', (string) $name)) {
             throw new Exception\InvalidArgumentException(sprintf(
                 '"%s" is not valid header name',
-                $name
+                (string) $name
             ));
         }
     }
--- a/src/HeaderSecurity.php
+++ b/src/HeaderSecurity.php
@@ -6,6 +6,7 @@ namespace Laminas\Diactoros;
 
 use function gettype;
 use function in_array;
+use function is_int;
 use function is_numeric;
 use function is_object;
 use function is_string;
@@ -148,16 +149,19 @@ final class HeaderSecurity
      */
     public static function assertValidName(mixed $name): void
     {
-        if (! is_string($name) && ! is_numeric($name)) {
+        if (! is_string($name) && ! is_int($name)) {
             throw new Exception\InvalidArgumentException(sprintf(
-                'Invalid header name type; expected string or numeric value; received %s',
+                'Invalid header name type; expected string or integer value; received %s',
                 is_object($name) ? $name::class : gettype($name)
             ));
         }
-        if (! preg_match('/^[a-zA-Z0-9\'`#$%&*+.^_|~!-]+$/D', (string) $name)) {
+
+        $name = is_int($name) ? (string) $name : $name;
+
+        if (! preg_match('/^[a-zA-Z0-9\'`#$%&*+.^_|~!-]+$/D', $name)) {
             throw new Exception\InvalidArgumentException(sprintf(
                 '"%s" is not valid header name',
-                (string) $name
+                $name
             ));
         }
     }
--- a/src/MessageTrait.php
+++ b/src/MessageTrait.php
@@ -7,9 +7,11 @@ namespace Laminas\Diactoros;
 use Psr\Http\Message\MessageInterface;
 use Psr\Http\Message\StreamInterface;
 
+use function array_is_list;
 use function array_map;
 use function array_merge;
 use function array_values;
+use function function_exists;
 use function implode;
 use function is_array;
 use function is_resource;
@@ -327,6 +329,10 @@ trait MessageTrait
      */
     private function setHeaders(array $originalHeaders): void
     {
+        if ($this->arrayIsList($originalHeaders)) {
+            throw new Exception\InvalidArgumentException('Header array must be an associative array');
+        }
+
         $headerNames = $headers = [];
 
         foreach ($originalHeaders as $header => $value) {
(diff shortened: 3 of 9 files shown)


src/HeaderSecurity.php
```diff
@@ -148,16 +148,16 @@ public static function assertValid(mixed $value): void
      */
     public static function assertValidName(mixed $name): void
     {
-        if (! is_string($name)) {
+        if (! is_string($name) && ! is_numeric($name)) {
```
Ocramius: I would expect them to be `string` anyway: casting a `float` to `string` is not our business, and may lead to trouble down the road.

Changing the regex: sure.

Casting `float` to a `string` is IMO a problem.

Ocramius (changes requested):

Xerkus:
I propose directly opposite: ban numeric headers as they are changing types and they *will* cause type errors somewhere unless name is always cast to string at every single place header name is obtained from array key.

Psalm will complain about casts, users will never do it. I know of this issue and I never do it.
The error is not of the kind that can be captured in developemnt or in QA because noone sane will ever try to inject numeric header in every single place headers are used in the codebase.

Xerkus:
I would like to call attention again to https://3v4l.org/XUj6h showcasing the issue

src/HeaderSecurity.php
Xerkus: Only integer like values are affected. Negative and positive. Integers that start with 0 are not converted, except singular 0.

https://3v4l.org/ei8MB
```
array(6) {
  [1]=>
  string(0) ""
  [-1]=>
  string(0) ""
  [0]=>
  string(0) ""
  ["0.1"]=>
  string(0) ""
  ["0123"]=>
  string(0) ""
  ["00"]=>
  string(0) ""
}
```

Ocramius: The problem is with casts that can also lead to a scientific notation 

Thinking https://stackoverflow.com/q/1471674/347063

Xerkus:
Type error places with numeric headers present (needs string cast):
https://github.com/laminas/laminas-diactoros/blob/df5279af2319f4ab3b2cf66f2701f9aada9968fe/src/AbstractSerializer.php#L144
https://github.com/laminas/laminas-diactoros/blob/df5279af2319f4ab3b2cf66f2701f9aada9968fe/src/MessageTrait.php#L337
https://github.com/laminas/laminas-diactoros/blob/df5279af2319f4ab3b2cf66f2701f9aada9968fe/src/RequestTrait.php#L265

Xerkus:
Reflect potential int
```diff
-     * @psalm-return array<non-empty-string, list<string>>
+     * @psalm-return array<non-empty-string|int, list<string>>

```
https://github.com/laminas/laminas-diactoros/blob/b5f2b9e9bb25b7e6ccf9208915f384a85f0eb540/src/MessageTrait.php#L105-L110

src/HeaderSecurity.php
weierophinney: The problem here is as follows:

- `ServerRequestFactory` calls on `Lamins\Diactoros\marshalHeadersFromSapi()`.
- `Laminas\Diactoros\marshalHeadersFromSapi()` loops through the `$_SERVER` array, looking for keys that match one of:
    - `REDIRECT_`
    - `CONTENT_`
    - `HTTP_`
- Keys that are of interest are then added to an array of discovered headers and returned to the factory.
- The factory uses the discovered headers to construct a `ServerRequest` instance
- `ServerRequest` takes that list of headers and passes them to `HeaderSecurity` to validate the names and values.

And this is where the issue comes in, because _PHP converts keys that are strings containing only integers to integers_. This happens at the point that `marshalHeadersFromSapi()` returns the header list, so once the `ServerRequest` starts to inject them into itself, they then fail validations.

I'm fine with not casting float to string, and can make that change, but the integer values are the ones that this patch is really concerned about. If I make the change to remove support for floats, will this resolve your reservations, @Ocramius ?

Xerkus: PHP only casts pure integer strings to int. Different kinds of notations like octal and such are not. So only `-?[0-9]+` is of concern. Anything else is not allowed.

weierophinney:
I've now updated to ensure we only consider integer values, and I've also put in some protections to ensure that non-assoc arrays cannot be used to populate the header list (e.g., via the `ServerRequest` factory).

src/MessageTrait.php
```diff
@@ -342,6 +350,27 @@ private function setHeaders(array $originalHeaders): void
...
+     *     array_is_list once minimum PHP version is 8.1.
+     */
+    private function arrayIsList(array $array): bool
+    {
+        if (function_exists('array_is_list')) {
+            return ! empty($array) && array_is_list($array);
+        }
+
+        $i = -1;
+        foreach ($array as $k => $v) {
```
Xerkus, line 368: will fail for empty array

weierophinney: Yes - for the same reason `array_as_list()` returns true for an empty array as well. Added a gate to this method.

weierophinney:
For reviewers: I can totally see why this might be considered a potentially bad idea, due to how PHP munges integer and string-integer keys, and the potential for accidentally setting the headers from a list versus an associative array. However, please note that `MessageTrait::setHeaders()` now explicitly looks for a non-empty _list_, and considers that invalid.

Second, there is a very real concern here. The following is a perfectly valid HTTP request:

```http
GET /some/path HTTP/1.1
Host: example.org
1: some value
Accept: application/json
```

However, currently, this would lead to a 500 with Diactoros because of that integer key... which RFC 7230 clearly allows in its header name ABNF.

With the protections against accepting lists, I think the changes here make sense, and definitely solve the issue reported in #11.

src/AbstractSerializer.php
```diff
@@ -137,11 +138,12 @@ protected static function serializeHeaders(array $headers): string
     /**
      * Filter a header name to wordcase
      *
-     * @param string $header
+     * @param string|int $header
      */
     protected static function filterHeader($header): string
     {
-        $filtered = str_replace('-', ' ', $header);
+        $filtered = is_int($header) ? (string) $header : $header;
```
boesing, line 145: would just cast always to string, since we have to check `is_int`, directly casting int and string might be even "faster".

src/MessageTrait.php
```diff
@@ -342,6 +350,31 @@ private function setHeaders(array $originalHeaders): void
...
+     * @todo Replace this method with a call to ! empty($array) &&
+     *     array_is_list once minimum PHP version is 8.1.
+     */
+    private function arrayIsList(array $array): bool
+    {
+        if (empty($array)) {
+            return false;
+        }
+
+        if (function_exists('array_is_list')) {
```
boesing, line 363: can't we just drop PHP <8.1? 
Did that in laminas-servicemanager for v4 as well, already seen other components dropping 8.0 so I guess that would fit?

Ocramius:
I'm OK with restricting this to integers, for now

src/MessageTrait.php
```diff
@@ -327,13 +330,18 @@ private function getStream($stream, string $modeIfNotInstance): StreamInterface
      */
     private function setHeaders(array $originalHeaders): void
     {
+        if ($this->arrayIsList($originalHeaders)) {
```
boesing, line 333: Do we need this check?

Is `0` disallowed in the RFC mentioned in the comments or why are numeric-only headers not allowed?
and if `0` is disallowed, why is `-1`, `0`, `1` allowed (which is not a list but still containers only numeric keys)?

boesing (changes requested):
Not 100% sure if the `array_is_list` check makes sense.

This throws no exception:

```php
[-1 => 'foo', 0 => 'bar', 1 => 'baz']
```

This throws exception:

```php
[0 => 'bar', 1 => 'baz']
```

But maybe I missed some detail.

----

Is there any plan to raise this via FIG?
Might be even worth again to ping this to PHP via RFC, I have had the [same problem](https://github.com/boesing/typed-arrays/blob/3f1b11bf25c7eab478079d63831eada014b8cf5f/src/OrderedList.php#L164-L186) with my typed arrays library and thus, I'd love to see an RFC which actually enforces to keep key-type. Would perfectly fit in PHP 9 to enforce.

src/MessageTrait.php
weierophinney: This is based on a polyfill for `array_is_list()` that mimics the behavior of that function... which only returns true if the keys are integers, are in sequence, and the sequence starts with 0. As such, keys starting at -1 will cause that function to return `false`.

All that said... I've noted on the Slack my reservations about the approach in this patch, and have asked for some guidance on an alternate approach.

weierophinney:
I'm going to withdraw this patch, and propose an alternate solution based on discussions in the Laminas Slack.

PR #154, opened 2023-05-02
Remove PhpInputStream

### LINKED ISSUE #150: PhpInputStream::getContent() inconsistency
Current implementation of `PhpInputStream::getContent()` accepts optional length parameter that makes `getContent()` to act similar to `read()`

Potentially multiple invocations are reading from php input stream, append to cache property and returning read chunk. This behavior changes, however, once whole of the php input stream is read into cache property: length argument is entirely ignored and `getContent()` returns full content of the cache property instead. 

Per PSR-7 the expected behavior is `Returns the remaining contents in a string`. Since php input stream is not rewindable once eof is reached `getContent()` behavior should be to always return empty string instead.

Only `__toString()` method should utilize cache property at this point to return whole body content.

weierophinney:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | yes
| New Feature   | no
| RFC           | no
| QA            | no

### Description

The main `PhpInputStream` existed was because prior to PHP 5.6, `php://input` was read-once and non-seekable.
As of 5.6, it's reusable AND seekable, which means there's no reason to keep the special implementation, particularly as it causes bugs (see #150)

This patch:

- Removes the class
- Updates the `ServerRequest` to, when a body of `php://input` is detected, create a read-only `Stream` instance with that value.

Fixes #150


Xerkus:
Thank you.

PR #155, opened 2023-05-03
Ensure fromGlobals does not use superglobals when passed empty arrays

### LINKED ISSUE #12: `ServerRequestFactory::fromGlobals()` checks if args are truthy rather than set
In `ServerRequestFactory::fromGlobals()`, most of the code checks if the arguments are truthy, rather than set. For example:

https://github.com/zendframework/zend-diactoros/blob/89d471cc09850c7d47839cf16ba7a1fed54d45e5/src/ServerRequestFactory.php#L57-L58

https://github.com/zendframework/zend-diactoros/blob/89d471cc09850c7d47839cf16ba7a1fed54d45e5/src/ServerRequestFactory.php#L72-L74

This becomes an issue during testing (or a long-running process that accepts multiple requests). For example:

```php
// In a previous test
$_POST = ['foo' => 'bar'];

// In current test
$request = ServerRequestFactory::fromGlobals(null, null, []);
echo serialize($request->getParsedBody());
// Expected: a:0:{}
// Actual: a:1:{s:3:"foo";s:3:"bar";}
```

Improved versions would be:

```php
        $server  = static::normalizeServer($server ?? $_SERVER); 
        $files   = static::normalizeFiles($files ?? $_FILES); 
```
```php
            $cookies ?? $_COOKIE, 
            $query ?? $_GET, 
            $body ?? $_POST, 
```

Or for under PHP 7:

```php
        $server  = static::normalizeServer(isset($server) ? $server : $_SERVER); 
        $files   = static::normalizeFiles(isset($files) ? $files : $_FILES); 
```
```php
            isset($cookies) ? $cookies : $_COOKIE, 
            isset($query) ? $query : $_GET, 
            isset($body) ? $body : $_POST, 
```

(This code should fix the issues, but I didn't have time to write any tests right now, so I'm opening an issue instead of a pr.)

---

Originally posted by @0b10011 at https://github.com/zendframework/zend-diactoros/issues/281

Xerkus:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | yes
| New Feature   | no
| RFC           | no
| QA            | no

### Description

`ServerRequestFactory::fromGlobals()` currently falls back to superglobals on null or when empty array is passed as an argument.

This change ensures fallback to superglobals only happens with null.

### DIFF
--- a/docs/book/v3/migration.md
+++ b/docs/book/v3/migration.md
@@ -1,5 +1,14 @@
 # Migration to Version 3
 
+## Changed
+
+The following features were changed in version 3.
+
+### ServerRequestFactory::fromGlobals
+
+The factory `Laminas\Diactoros\ServerRequestFactory::fromGlobals()` was modified such that passing empty array values for arguments that accept `null` or an array now will not use the associated superglobal in that scenario.
+Previously, an empty array value was treated as identical to `null`, and would cause the factory to fallback to superglobals; now, this is a way to provide an empty set for the associated value(s).
+
 ## Removed
 
 The following features were removed for version 3.
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -185,20 +185,6 @@
     </DocblockTypeContradiction>
   </file>
   <file src="src/ServerRequestFactory.php">
-    <LessSpecificReturnStatement>
-      <code><![CDATA[$requestFilter(new ServerRequest(
-            $server,
-            $files,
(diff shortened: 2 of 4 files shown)


weierophinney:
@Xerkus I've added some migration documentation for this change, and will merge if/when CI passes (your changes already passed; just need to validate documentation linting now).

PR #157, opened 2023-05-04
Filter integer header names during SAPI discovery

### LINKED ISSUE #11: Handle request headers with numeric keys

weierophinney:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | yes
| BC Break      | yes
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Integer strings have the nasty habit of being cast to actual integers by PHP, making them problematic for usage in an associative array, despite being valid per the RFC 7230 ABNF.

Additionally, having them pass through `marshal_headers_from_sapi()` means that once `ServerRequest` gets them and tries to use them, `HeaderSecurity::assertValidName()` will raise an exception for integers, which could lead to unexpected server errors.

This patch chooses to filter such header names out entirely.
It also documents the change, and ways to address it if you previously depended on integer header field names.

Fixes #11


src/functions/marshal_headers_from_sapi.php
```diff
@@ -63,5 +66,6 @@ function marshalHeadersFromSapi(array $server): array
         }
     }
 
-    return $headers;
+    // Filter out integer keys
+    return array_filter($headers, fn(string|int $key): bool => is_string($key), ARRAY_FILTER_USE_KEY);
```
Xerkus: Please document inline where the integer keys come from.

Xerkus (commented):
`ServerRequestFactory::fromGlobals()` needs test to ensure it filters out integer-like headers since `marshalHeadersFromSapi()` is an implementation detail.

docs/book/v3/migration.md
```diff
@@ -24,6 +24,12 @@ For consumers, usage should be completely backwards compatible, however.
 The factory `Laminas\Diactoros\ServerRequestFactory::fromGlobals()` was modified such that passing empty array values for arguments that accept `null` or an array now will not use the associated superglobal in that scenario.
 Previously, an empty array value was treated as identical to `null`, and would cause the factory to fallback to superglobals; now, this is a way to provide an empty set for the associated value(s).
 
+### marshalHeadersFromSapi
+
+The function `Laminas\Diactoros\marshalHeadersFromSapi()`, which is consumed by `Laminas\Diactoros\ServerRequestFactory::fromGlobals()`, was modified such that it will no filter out header field names that evaluate to integers.
```
Xerkus: typo
```suggestion
The function `Laminas\Diactoros\marshalHeadersFromSapi()`, which is consumed by `Laminas\Diactoros\ServerRequestFactory::fromGlobals()`, was modified such that it will now filter out header field names that evaluate to integers.
```

docs/book/v3/security.md
```diff
@@ -0,0 +1,39 @@
...
+
+[PSR-7](https://www.php-fig.org/psr/psr-7/) targets [RFC 7230](https://www.rfc-editor.org/rfc/rfc7230).
+RFC-7230 defines an ABNF pattern for header field names that allows the possibility of using an integer as a header field; e.g.,
+
+```http
+1234: header value
+```
+
+The PSR-7, `Psr\Http\MessageInterface::getHeaders()` method requires implementations to return an associative array, where the key is the header field name.
+This triggers an interesting quirk in PHP: when adding a value to an array using a string that consists of an integer value, PHP will convert this value to an integer.
```
Xerkus: Should this link to original php bug report?
https://bugs.php.net/bug.php?id=80309

docs/book/v3/security.md
```diff
@@ -0,0 +1,39 @@
...
+This presents several issues:
+
+- First, it means that consumers cannot depend on the header field name returned being a string.
+- Second, our own validation of header field name will fail, as it will not see a string.
+
+Normally, this will not present an issue, as the way to add headers to a message is via the `MessageInterface::withHeader()` and `MessageInterface::withAddedHeader()` methods, which both require a `string` name argument.
```
Xerkus, line 35: note: `string` on name argument does not prevent header from being added as `'1234'`. It will definitely present an issue, just not immediately as on PSR-7 message creation where it had chance to convert to int before validation.

Xerkus (changes requested):
A typo, otherwise good.

docs/book/v3/security.md
weierophinney: Correct. 

In fact... in that case, `getHeaders()` will still return the key as well. Which means there's no guarantee on the consumer end that the array from that has only string keys.

I'll add a note to that effect.

src/functions/marshal_headers_from_sapi.php
```diff
@@ -63,5 +66,9 @@ function marshalHeadersFromSapi(array $server): array
         }
     }
 
-    return $headers;
+    // Filter out integer keys.
+    // These can occor if the translated header name is a string integer.
+    // PHP will cast those to integers when assigned to an array.
+    // This filters them out.
```
Xerkus: Typo

```suggestion
    // These can occur if the translated header name is a string integer.
    // PHP will cast those to integers when assigned to an array.
    // This filters them out.
```

Xerkus (approved):

PR #158, opened 2023-05-05
Fix docs referring to array key as value

Xerkus:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description
Problematic behavior is exhibited by the array key but documentation implies the value will be converted instead.


Ocramius:
Thanks @Xerkus!

PR #163, opened 2023-06-02
Add RectorPHP

internalsystemerror:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | yes
| QA            | yes

### Description

I thought I'd take a look at how we might use https://github.com/rectorphp/rector. This PR contains 2 commits to first add rector as a dependency with configuration, and then to actually run rector and see what has changed.

Its not my intention for this to merged as is, but as a demonstration to get feedback and help discuss an approach.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,6 +48,7 @@
         "php-http/psr7-integration-tests": "^1.3",
         "phpunit/phpunit": "^9.5.28",
         "psalm/plugin-phpunit": "^0.18.4",
+        "rector/rector": "^0.17.0",
         "vimeo/psalm": "^5.9"
     },
     "provide": {
--- a/rector.php
+++ b/rector.php
@@ -1,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+use Rector\Config\RectorConfig;
+use Rector\Set\ValueObject\LevelSetList;
+
+return static function (RectorConfig $rectorConfig): void {
+    // Set paths
+    $rectorConfig->paths([
+        __DIR__ . '/src',
+        __DIR__ . '/test',
+    ]);
+
+    // Define set list to upgrade PHP
+    $rectorConfig->sets([
+        LevelSetList::UP_TO_PHP_80,
+    ]);
+};
--- a/test/Response/JsonResponseTest.php
+++ b/test/Response/JsonResponseTest.php
@@ -19,6 +19,7 @@ use const JSON_HEX_APOS;
 use const JSON_HEX_QUOT;
 use const JSON_HEX_TAG;
 use const JSON_PRETTY_PRINT;
+use const JSON_THROW_ON_ERROR;
 use const JSON_UNESCAPED_SLASHES;
 
 class JsonResponseTest extends TestCase
@@ -139,7 +140,7 @@ class JsonResponseTest extends TestCase
         $json     = ['test' => 'data'];
         $response = new JsonResponse($json);
 
-        $actual = json_decode($response->getBody()->getContents(), true);
+        $actual = json_decode($response->getBody()->getContents(), true, 512, JSON_THROW_ON_ERROR);
         $this->assertSame($json, $actual);
     }
 
@@ -158,7 +159,7 @@ class JsonResponseTest extends TestCase
         $this->assertNotSame($response, $newResponse);
 
         $this->assertSame($json, $newResponse->getPayload());
-        $decodedBody = json_decode($newResponse->getBody()->getContents(), true);
+        $decodedBody = json_decode($newResponse->getBody()->getContents(), true, 512, JSON_THROW_ON_ERROR);
         $this->assertSame($json, $decodedBody);
     }
 
--- a/.gitattributes
+++ b/.gitattributes
@@ -10,6 +10,7 @@
 /phpunit.xml.dist export-ignore
 /psalm-baseline.xml export-ignore
 /psalm.xml.dist export-ignore
+/rector.php export-ignore
 /renovate.json export-ignore
 /test/ export-ignore
 /psalm/ export-ignore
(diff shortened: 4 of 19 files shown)


test/Response/JsonResponseTest.php
```diff
@@ -139,7 +140,7 @@ public function testConstructorRewindsBodyStream(): void
         $json     = ['test' => 'data'];
         $response = new JsonResponse($json);
 
-        $actual = json_decode($response->getBody()->getContents(), true);
+        $actual = json_decode($response->getBody()->getContents(), true, 512, JSON_THROW_ON_ERROR);
```
weierophinney, line 143: This is the first change I'm not entirely clear on. I cannot remember what that third argument is supposed to be so seeing the 512 value makes me scratch my head. That said having an exception here makes a lot of sense since it will ensure that the assertion never gets run. I wonder if it can use named arguments here instead? Perhaps @samsonasik can answer that.

weierophinney (commented):
Overall this is very interesting as it highlights a number of changes that are easy to overlook as we bump versions.

test/Response/JsonResponseTest.php
Ocramius: I think it would be best to use `flags: JSON_THROW_ON_ERROR` here

composer.json
```diff
@@ -48,6 +48,7 @@
         "php-http/psr7-integration-tests": "^1.3",
         "phpunit/phpunit": "^9.5.28",
         "psalm/plugin-phpunit": "^0.18.4",
+        "rector/rector": "^0.17.0",
```
Ocramius, line 51: This is something I'd really like to see in our CI, rather than in a dev-dependency.

That said, it is fine here too, **for now**

test/Response/JsonResponseTest.php
internalsystemerror: 512 is the default value. Named parameters were introduced in PHP 8.0, so curious why it didn't use them for this. I guess this demonstrates that rector may be a start but it will nearly always required manual intervention?

composer.json
internalsystemerror: There is no phar available for rector, and we could include it as a global dependency in CI, but that could lead to an introduced failure without any code changes in the repositories.

Ocramius: @internalsystemerror I would make a separate `composer.lock` just for that, installed globally in our CI container.

Not a problem for now though: just raising the idea :-)

Related: https://github.com/laminas/laminas-ci-matrix-action/issues/131

test/Response/JsonResponseTest.php
Ocramius: Yes, it certainly always needs human review

composer.json
internalsystemerror: No please keep it coming, thats the point of this PR as a draft, to get feedback!

test/Response/JsonResponseTest.php
samsonasik: I think `JSON_THROW_ON_ERROR` is fine here as this is test class

samsonasik: named arg currently used on rector for attribute creation

composer.json
samsonasik: Use pinned version 0.17.0 is better imo

rector.php
```diff
@@ -0,0 +1,19 @@
...
+    $rectorConfig->paths([
+        __DIR__ . '/src',
+        __DIR__ . '/test',
+    ]);
+
+    // Define set list to upgrade PHP
+    $rectorConfig->sets([
+        LevelSetList::UP_TO_PHP_80,
+    ]);
+};
```
samsonasik, line 19: This is my current config for some laminas components I run:

```php
<?php

declare(strict_types=1);

use Rector\CodingStyle\Rector\ArrowFunction\StaticArrowFunctionRector;
use Rector\CodingStyle\Rector\Closure\StaticClosureRector;
use Rector\Config\RectorConfig;
use Rector\Php55\Rector\String_\StringClassNameToClassConstantRector;
use Rector\Php56\Rector\FunctionLike\AddDefaultValueForUndefinedVariableRector;
use Rector\Php70\Rector\FuncCall\RandomFunctionRector;
use Rector\Php71\Rector\FuncCall\CountOnNullRector;
use Rector\Php73\Rector\FuncCall\JsonThrowOnErrorRector;
use Rector\Privatization\Rector\Property\PrivatizeFinalClassPropertyRector;
use Rector\Set\ValueObject\LevelSetList;
use Rector\TypeDeclaration\Rector\ArrowFunction\AddArrowFunctionReturnTypeRector;
use Rector\TypeDeclaration\Rector\Closure\AddClosureReturnTypeRector;

return static function (RectorConfig $rectorConfig): void {
    $rectorConfig->sets([
        LevelSetList::UP_TO_PHP_80,
    ]);

    $rectorConfig->rules([
        StaticArrowFunctionRector::class,
        StaticClosureRector::class,
        AddClosureReturnTypeRector::class,
        PrivatizeFinalClassPropertyRector::class,
        AddArrowFunctionReturnTypeRector::class,
    ]);

    $rectorConfig->parallel();
    $rectorConfig->paths([
        __DIR__ . '/src',
        __DIR__ . '/test',
    ]);

    $rectorConfig->skip([
        // possibly too detail on some cases?
        CountOnNullRector::class,

        // possibly null undefined on purpose?
        AddDefaultValueForUndefinedVariableRector::class,

        // define list of services?
        StringClassNameToClassConstantRector::class,

        // not a secure lib on purpose?
        RandomFunctionRector::class,

        // probably handle after execute?
        JsonThrowOnErrorRector::class =>  [
            __DIR__ . '/src',
        ],
    ]);
};

```

PR #169, opened 2023-08-01
Correctly handle explicitly providing `JSON_THROW_ON_ERROR` in `JsonResponse`

TimWolla:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | maybe
| New Feature   | no
| RFC           | no
| QA            | no

### Description

If the `JSON_THROW_ON_ERROR` flag is explicitly given, `jsonEncode()` violates its contract to throw `InvalidArgumentException` if it is unable to encode the input as JSON.

### DIFF
--- a/src/Response/JsonResponse.php
+++ b/src/Response/JsonResponse.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Laminas\Diactoros\Response;
 
+use JsonException;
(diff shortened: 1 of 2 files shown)


Ocramius (approved):
Good improvement, thanks @TimWolla!

PR #174, opened 2023-10-23
Remove support for PHP 8.0, remove unnecessary reflection method calls

gsteel:
### Description

Just investigating upgrading to PHPUnit 10 here to find out that it's still blocked by the interop integration tests.

Considering that we'll be dropping 8.0 for PHPUnit 10 anyhow, this patch just chips off one job from that upgrade.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         },
         "sort-packages": true,
         "platform": {
-            "php": "8.0.99"
+            "php": "8.1.99"
         }
     },
     "extra": {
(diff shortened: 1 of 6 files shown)


froschdesign (approved):

gsteel:
@froschdesign

I don't think it's necessarily worth releasing this, what do you think?

froschdesign:
@gsteel 
You have planned further changes, so it is possible to summarise this.

Ocramius:
I'm releasing this as-is: any further improvements can safely land in `3.4.0` later.

Issue #175, opened 2023-10-26
2.x series does not support PHP 8.3


dereuromark:
#### Summary

PHP 8.3 support is needed for 2.x series.

See
https://github.com/cakephp/cakephp/issues/17387

#### Current behavior

Not installable, probably for no good reason as 8.3 should be compatible out of the box

#### How to reproduce

See above link

#### Expected behavior

~8.3 included in the constraint for PHP so folks stuck on that 2.x series of this library can still do 8.3 checking and development of their library code.

Many libraries and frameworks will have to support their major versions for a few years longer, so it is important that they can ship with secure/safe PHP versions in the present and future.
As per semver they cannot just modify their public API and as such are locked within 2.x series for now (no types added etc).

Thank you

Ocramius:
I don't think `2.x` should be revived: is the adoption (either bump or dependency widening) to `3.x` not viable?

Can we perhaps help there instead, if there are any blockers?

froschdesign:
@dereuromark 
> It shows that laminas/laminas-diactoros in all compatible versions to cake 4.4 and 4.5 do not offer 8.3 support due to their bad way of constraining.

What do you mean with "bad way of constraining"?

Ocramius:
From your upstream issue:

```

    - cakephp/cakephp[4.4.17, ..., 4.5.0] require laminas/laminas-diactoros ^2.2.2 -> satisfiable by laminas/laminas-diactoros[2.2.2, ..., 2.25.2].
```

This is what I'd suggest widening to `"laminas/laminas-diactoros": "^2 || ^3"` first

Ocramius:
> @dereuromark
> 
> > It shows that laminas/laminas-diactoros in all compatible versions to cake 4.4 and 4.5 do not offer 8.3 support due to their bad way of constraining.
> 
> What do you mean with "bad way of constraining"?

See https://github.com/laminas/laminas-diactoros/issues/117

It is a past (and present) disagreement from the issue author with our approach to dependency management, which we explained in depth to be well founded, well intended, and I assume well understood now.

Ocramius:
BTW, we use `--ignore-platform-reqs` in some CI pipelines ourselves, where upgrades aren't feasible yet :-)

dereuromark:
> I don't think 2.x should be revived: is the adoption (either bump or dependency widening) to 3.x not viable?

https://github.com/cakephp/cakephp/pull/17388 shows that this is not possible.

```
[...]

PR #176, opened 2023-10-27
Update PHP constraints to allow v8.3

ADmad:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | yes
| QA            | no

### Description

Refs #175


ADmad:
The "QA Checks' failure seems to be due to a bug in psalm which is addressed here https://github.com/vimeo/psalm/pull/10186 but hasn't been released yet.

composer.lock
```diff
@@ -3998,16 +3999,16 @@
         },
         {
             "name": "vimeo/psalm",
-            "version": "5.10.0",
+            "version": "5.15.0",
```
Ocramius: Let's not upgrade anything: just the lock file.

Merging up is hell otherwise.

Ocramius (changes requested):

composer.lock
Ocramius: Also, let's squash in the change that reverts all these lockfile changes :)

Ocramius (approved):
Thanks @ADmad!

ADmad:
I reverted the changes and then ran `compose update --lock`, not sure why it downgraded the version for `plugin-api-version`.

ADmad:
> I reverted the changes and then ran compose update --lock, not sure why it downgraded the version for plugin-api-version.

TIL: The value for `plugin-api-version` in `composer.lock` depends on the composer version used. I had used an older `2.0.4` version of composer hence `plugin-api-version` was set to `2.0.0`.

PR #178, opened 2023-11-10
Cover invalid arguments to Stream constructor

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### DIFF
--- a/test/StreamTest.php
+++ b/test/StreamTest.php
@@ -5,9 +5,11 @@ declare(strict_types=1);
 namespace LaminasTest\Diactoros;
 
 use CurlHandle;
+use Exception;
 use GdImage;
 use InvalidArgumentException;
 use Laminas\Diactoros\Exception\InvalidArgumentException as DiactorosInvalidArgumentException;
+use Laminas\Diactoros\Exception\RuntimeException as DiactorosRuntimeException;
 use Laminas\Diactoros\Stream;
 use PHPUnit\Framework\TestCase;
 use ReflectionProperty;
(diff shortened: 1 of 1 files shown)


test/StreamTest.php
```diff
@@ -68,6 +73,28 @@ public function testCanInstantiateWithStreamResource(): void
...
+    public function testInvalidStringArgumentWithCustomErrorHandler(): void
+    {
+        set_error_handler(function (): never {
+            throw new Exception('Bad News');
+        }, E_WARNING);
+
+        try {
+            new Stream('foo');
+        } catch (DiactorosRuntimeException $error) {
+            self::assertStringContainsString('Invalid stream reference provided', $error->getMessage());
```
lcobucci, line 92: Shouldn't we expect to see "Bad news" here?

Ocramius: Probably previous exception? 🤔

Xerkus:
What exactly is this supposed to achieve? 
`fopen()` failing to open stream is covered by further checks.

This is in response to a question in our slack about `fopen()` with url that gives 404 where false is returned and warning is triggered, right?
Expecting some outside error handler to convert warnings to exceptions is not normal and should not be done in tests. To deal with it properly error handler should be added around `fopen()` call.

gsteel:
@Xerkus - Yes, I added these tests last night after talking to @ramsey on Slack. I couldn't find any tests that checked for that specific exception. Whether it's valid or not to consider user defined error handlers, I'll leave up to you. Feel free to merge or close.

Adding an error handler to `/src/Stream` to ensure that warnings from `fopen` are _not_ suppressed and _are_ correctly raised as `Throwable` would probably be a worthwhile addition.

test/StreamTest.php
gsteel: 'Bad News' is swallowed by `src/Stream` try/catch - it should be there in `$error->previous`

Issue #159, opened 2023-05-05
Numeric header names handling in PSR-7 message objects


Xerkus:
PHP has little known quirk where array converts string keys consisting of integer value to integers . See [PHP bug 80309](https://bugs.php.net/bug.php?id=80309) for more details.

Numeric header names failing type validation during `ServerRequest` instance creation from SAPI environment were reported in #11 and #157 handled it by filtering out numeric headers during server request creation from PHP globals.

However such headers could still be added to PSR-7 messages via `withAddedHeader(string $name)` and `withHeader(string $name)` methods, for example using string name like `'1234'`. Those headers will be an integer type when returned from `getHeaders()`, potentially causing ripple effect across user codebase expecting strongly typed string argument.

Since this behavior is universal across all implementations, @weierophinney will work with PHP-FIG to provide PSR-7 errata detailing expected handling of such headers. This is a tracking issue until such resolution is available.

PR #190, opened 2024-08-01
Upgrade PHPUnit to 10.x. Fix Stream conformance

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -379,6 +379,14 @@
       <code><![CDATA[RequestInterfaceStaticReturnTypes]]></code>
     </UnusedClass>
   </file>
+  <file src="test/StreamTest.php">
+    <PossiblyUnusedMethod>
+      <code><![CDATA[invalidResources]]></code>
+      <code><![CDATA[invalidStringResources]]></code>
+      <code><![CDATA[provideDataForIsReadable]]></code>
+      <code><![CDATA[provideDataForIsWritable]]></code>
+    </PossiblyUnusedMethod>
+  </file>
   <file src="test/TestAsset/CallbacksForCallbackStream.php">
     <PossiblyUnusedMethod>
       <code><![CDATA[sampleCallback]]></code>
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,22 +1,22 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="5.17.0@c620f6e80d0abfca532b00bda366062aaedf6e5d">
+<files psalm-version="5.25.0@01a8eb06b9e9cc6cfb6a320bf9fb14331919d505">
   <file src="src/AbstractSerializer.php">
     <RedundantCondition>
-      <code>! $crFound</code>
-      <code>! $crFound</code>
+      <code><![CDATA[! $crFound]]></code>
+      <code><![CDATA[! $crFound]]></code>
     </RedundantCondition>
   </file>
   <file src="src/CallbackStream.php">
     <ImplementedReturnTypeMismatch>
-      <code>null|callable</code>
+      <code><![CDATA[null|callable]]></code>
     </ImplementedReturnTypeMismatch>
     <MixedAssignment>
-      <code>$contents</code>
+      <code><![CDATA[$contents]]></code>
     </MixedAssignment>
     <PossiblyUnusedReturnValue>
-      <code>int</code>
-      <code>int</code>
-      <code>string</code>
+      <code><![CDATA[int]]></code>
+      <code><![CDATA[int]]></code>
+      <code><![CDATA[string]]></code>
     </PossiblyUnusedReturnValue>
   </file>
   <file src="src/Exception/DeserializationException.php">
@@ -27,12 +27,12 @@
   </file>
   <file src="src/Exception/InvalidProxyAddressException.php">
     <PossiblyUnusedMethod>
-      <code>forInvalidProxyArgument</code>
+      <code><![CDATA[forInvalidProxyArgument]]></code>
     </PossiblyUnusedMethod>
   </file>
   <file src="src/HeaderSecurity.php">
     <UnusedConstructor>
-      <code>__construct</code>
+      <code><![CDATA[__construct]]></code>
     </UnusedConstructor>
   </file>
   <file src="src/MessageTrait.php">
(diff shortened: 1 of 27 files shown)


gsteel:
This uses 1.x-dev of `php-http/psr7-integration-tests` in order to install PHPUnit 10.

There's no release there yet: https://github.com/php-http/psr7-integration-tests/issues/76

Some failing tests need attention…

froschdesign:
The release is here: https://github.com/php-http/psr7-integration-tests/releases/tag/1.4.0

gsteel:
So we can't upgrade to PHPUnit 10 without first resolving BC issues around numeric header names, and we won't be able to support PHP 8.4 without upgrading to PHPUnit 10.

This patch is a BC break (probably) because it casts header names to string prior to validation on the assumption that numeric keys have already been cast to int by PHP.

Paging @laminas/technical-steering-committee for input…

Do we need to release a major here?

internalsystemerror:
My thoughts are, BC break = Major version bump. What would be the burden/implications from this?

Xerkus:
Can you elaborate? Numeric headers can not be handled by the library interface and it can not be resolved.
See #159

weierophinney:
Agreed with @Xerkus - header names cannot be purely numeric (and this not integers). If anything, this feels like resolving a potential error for users, and not introducing a break.

gsteel:
It appears to me that tests no longer pass with the previous behaviour of throwing an exception for integer header names when creating the request from globals.

To get the integration tests to pass, I've cast numerics to string in the appropriate place, but this is the BC break - currently, users expect integer header names to be exceptional and this patch makes them valid as per RFC whatever.

It doesn't affect the existing issue that `withHeader('123')` will yield `int 123` from `getHeaders`

gsteel:
> Agreed with @Xerkus - header names cannot be purely numeric (and this not integers). If anything, this feels like resolving a potential error for users, and not introducing a break.

@weierophinney OK - maybe I'm a bit confused here - I thought that numeric header names were valid according to RFC 7230 ??

Xerkus:
Please see the issue I linked and issues it refers to. We can not allow numeric headers.

Xerkus:
Please remove type casting for headers from the PR. It hides the issue for the current test implementation.

gsteel:
@Xerkus - I'm going to go back to the drawing board on this patch to identify properly where existing test cases are failing.

Xerkus:
@gsteel for context, we filtered out numeric headers from globals when request is created but did not forbid numeric headers in the headers security check because @weierophinney wanted to bring it to PHP-FIG first.

There are numeric headers in the wild from a random http client as evidenced by the original report but we couldn't identify any valid use case for them. Strictly speaking, unregistered headers not prefixed with `x-` are not RFC conforming. Hence why we should not try to support them just because they fit ABNF.

gsteel:
Thanks @Xerkus and @weierophinney 

So it turns out that I added some failing tests around this during the upgrade to PHPUnit 10. Whilst I was waiting for the integration tests to get released, I forgot I'd added them and assumed they were new failures.

Generally speaking, I'm just a bit of an idiot!!

Sorry for wasting everyone's time - this should be a straight-forward review now.

akrabat:
> Strictly speaking, unregistered headers not prefixed with x- are not RFC conforming.

This is no longer true with RFC9110, where [Section 16.3.2.1](https://www.rfc-editor.org/rfc/rfc9110.html#section-16.3.2.1) states:

> Field names ought not be prefixed with "X-"

Numeric only field names are discouraged, but not illegal:

> While the field-name syntax is defined to allow any token character, in practice some implementations place limits on the characters they accept in field-names. To be interoperable, new field names **SHOULD** constrain themselves to alphanumeric characters, "-", and ".", and **SHOULD** begin with a letter. For example, the underscore ("_") character can be problematic when passed through non-HTTP gateway interfaces (see [Section 17.10](https://www.rfc-editor.org/rfc/rfc9110.html#underscore.in.fields)).[¶](https://www.rfc-editor.org/rfc/rfc9110.html#section-16.3.2.1-3)

Xerkus:
> This is no longer true with RFC9110, where [Section 16.3.2.1](https://www.rfc-editor.org/rfc/rfc9110.html#section-16.3.2.1) states:

I was not aware we got newer RFC on this. Thank you.

src/Stream.php
```diff
@@ -318,14 +318,18 @@ public function getMetadata(?string $key = null)
...
+        $error    = null;
         $resource = $stream;
 
         if (is_string($stream)) {
             try {
                 $resource = fopen($stream, $mode);
             } catch (Throwable $error) {
+            }
+
+            if (! is_resource($resource)) {
```
Xerkus, line 330: This is already handled by the next check with `isValidStreamResourceType`. This change should not be necessary.

gsteel: Different exception types are expected depending on the failure. Runtime exception is expected by interop integration tests when the file path does not exist for example

Xerkus: Then I think it should be two separate exceptions. `Invalid stream reference provided: ` does not make much sense. 

That try-catch block is pretty much only for the empty string throwing ValueError and may be for the mezzio error handler converting warnings to exceptions.
What about the warning emitted by fopen on non-existent path? There was error handler but it was removed for whatever reason. Does interop say something about that?

test/StreamTest.php
```diff
@@ -55,6 +58,14 @@ protected function tearDown(): void
...
             unlink($this->tmpnam);
         }
+
+        foreach (self::$tempFiles as $tempFile) {
+            if (! file_exists($tempFile)) {
+                continue;
+            }
+
+            unlink($tempFile);
+        }
```
Xerkus, line 73: This is wrong. those temp files are created by a provider which is executed before all tests and as such this must be cleaned after all tests.

src/Stream.php
gsteel: The error handler was removed in #103 - I'm unsure exactly why, but the effect is that `fopen` warnings are no longer silenced.

The try/catch does indeed only catch the empty string case.

Considering that the exception is thown for both empty string and non-existent path, the contract with Interop for throwing `\RuntimeException` is satisfied, and I think it's preferable to expose the PHP warnings to users rather than swallow them in the exception.

What do you suggest the change should be here? Re-introduce the error handler? Change the exception message?

Xerkus: It seems removing the warning to exception conversion was an unrelated change. Adding it back here would be an unrelated change again. 

I think exception message should be improved to include the offending string and it would be sufficient for now.

psalm-baseline.xml
```diff
@@ -1,22 +1,22 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="5.17.0@c620f6e80d0abfca532b00bda366062aaedf6e5d">
+<files psalm-version="5.25.0@01a8eb06b9e9cc6cfb6a320bf9fb14331919d505">
   <file src="src/AbstractSerializer.php">
     <RedundantCondition>
-      <code>! $crFound</code>
-      <code>! $crFound</code>
+      <code><![CDATA[! $crFound]]></code>
+      <code><![CDATA[! $crFound]]></code>
```
Xerkus, line 6: I wonder if this CDATA thrashing in the baseline is due to different xml libraries available when baseline is updated.

gsteel: Maybe - I have libxml 2.9.13 locally for PHP 8.1

src/Stream.php
```diff
@@ -318,14 +318,21 @@ public function getMetadata(?string $key = null)
...
                 throw new Exception\RuntimeException(
-                    sprintf('Invalid stream reference provided: %s', $error->getMessage()),
+                    sprintf(
+                        'Empty on non-existent stream identifier or file path provided: "%s"',
```
Xerkus: > empty on

A typo?

gsteel: 🤦‍♂️

Xerkus:
@gsteel Thank you

PR #193, opened 2024-09-10
Fix cookie parsing from headers in server request factory using globals.

### LINKED ISSUE #165: Plus signs in cookie data get converted to space.
### Bug Report

Plus signs in cookie values are converted to spaces.

|    Q        |   A
|------------ | ------
| Version(s)  | 2.24.0 but also the latest version.

My reading of [rfc6265](https://datatracker.ietf.org/doc/html/rfc6265) is that plus signs:

> cookie-octet      = %x21 / %x23-2B / %x2D-3A / %x3C-5B / %x5D-7E

as the plus sign is a %2B

#### Summary

Plus signs in cookie values are converted to spaces.

#### Current behavior

The function parseCookieHeader does a urldecode on the values [here](https://github.com/laminas/laminas-diactoros/blob/9abef5d1fd2f2cbe42032413a35eead833dbac63/src/functions/parse_cookie_header.php#L36), and so the plus sign is converted to a space.

#### How to reproduce

$data = "john=12345; plus_sign=plus+sign";
$result = parseCookieHeader($data);
var_dump($data, $result);

#### Expected behavior

As the plus sign is apparently a valid cookie character, it should be passed through unchanged.

Not sure how you could change this without a large BC break...

btw, probably relevant
https://bugs.php.net/bug.php?id=78929

Xerkus:
PHP used to encode space in cookie values as + in violation of RFCs until it was fixed in PHP 7.4.
See https://bugs.php.net/bug.php?id=79174

We decode headers to read server request cookies when creating server request from globals. It was still converting `+` in the cookie value back to space.

This PR fixes our implementation to align it again with the supported PHP versions and RFCs.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="5.25.0@01a8eb06b9e9cc6cfb6a320bf9fb14331919d505">
+<files psalm-version="5.26.0@4787eaf414e16c661902b94dfe5d882223e5b513">
   <file src="src/AbstractSerializer.php">
     <RedundantCondition>
       <code><![CDATA[! $crFound]]></code>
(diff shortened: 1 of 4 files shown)


gsteel (approved):
LGTM! My only concern here is that it's a behaviour change that is possibly unexpected, even though it's effectively a fix.

Can we get away with releasing this in a patch?

Xerkus:
@gsteel  You want to release it as 3.3.1?
We can do that but I was going to tag 3.4.0 anyway although phpunit 10 upgrade is a dev dependency and the whole thing can be easily reduced to a patch release.

gsteel:
I think a 3.4.0 release is fine - I was questioning the change needing to go into 4.0 as a possible BC break, or not as the case may be

Xerkus:
There is no chance for BC break here.

PR #194, opened 2024-09-10
Fix provided versions of psr/http-factory-implementation in composer.json

### LINKED ISSUE #161: `composer.json` provides non-existing versions of `psr/http-factory`
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 3.0.0

#### Summary

With v3, the `provide` section got updated. Somehow, `psr/http-factory-implementation` was also updated to v1.1 and v2.0 which do not exist.

Xerkus:
None of the provided versions of the virtual package psr/http-factory-implementation exist in psr/http-factory. Particularly there is no v2 version.

Original change was meant for the psr/http-message-implementation but got inadvertently duplicated.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,7 +51,7 @@
         "vimeo/psalm": "^5.25.0"
     },
     "provide": {
-        "psr/http-factory-implementation": "^1.1 || ^2.0",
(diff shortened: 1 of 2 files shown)


gsteel (approved):
👍

PR #196, opened 2024-09-11
Add basic host validation in Uri and add port validation to the Uri constructor

### LINKED ISSUE #172: `Uri::__toString()` can yield malformed URIs
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | Current git 3.3.x

#### Summary

The `Uri` class is able to parse malformed URIs but this results in `Uri::__toString()` generating a malformed URI. Attempting to pass that URI back into Uri will yield a `InvalidArgumentException`.

#### Current behavior

Certain malformed URIs do not round-trip through `\Laminas\Diactoros\Uri`.

#### How to reproduce

```php
<?php

require('vendor/autoload.php');

$value = 'http://invalid:%20https://example.com';
$uri = new \Laminas\Diactoros\Uri($value);
$uri2 = new \Laminas\Diactoros\Uri($uri->__toString()); // Exception is thrown here.
```

#### Expected behavior

Either both constructors throw, or neither.

---------------

Note: This issue also exists in guzzlehttp/psr7 and was reported at guzzle/psr7#583.

Xerkus:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | yes/no
| New Feature   | no
| RFC           | yes
| QA            | no

### Description

This change introduces validation for the uri host as `parse_url()` does not validate host subcomponent permitting outright forbidden characters to be used.

This PR introduces basic host subcomponent validation that would reject forbidden characters, validate IPv6 address and wrap it the square brackets. 
This validation does not ensure hostname is actually valid. For example, URI allows various sub-delimeters  and percent encoding which would be invalid in the context of the http urls. This validation does not prevent them and it does not validate percent encoding is actually valid.

See [RFC3986 3.2.2](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2)

Port validation was added to the Uri constructor as `parse_url()` would allow port 0 which is a wildcard port not valid for the Uri.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -408,6 +408,7 @@
     </MixedAssignment>
     <PossiblyUnusedMethod>
       <code><![CDATA[authorityInfo]]></code>
+      <code><![CDATA[invalidHosts]]></code>
       <code><![CDATA[invalidPaths]]></code>
       <code><![CDATA[invalidQueryStrings]]></code>
       <code><![CDATA[invalidUriProvider]]></code>
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -406,6 +406,19 @@
     <MixedAssignment>
       <code><![CDATA[$test]]></code>
     </MixedAssignment>
+    <PossiblyUnusedMethod>
+      <code><![CDATA[authorityInfo]]></code>
+      <code><![CDATA[invalidPaths]]></code>
+      <code><![CDATA[invalidQueryStrings]]></code>
+      <code><![CDATA[invalidUriProvider]]></code>
+      <code><![CDATA[mutations]]></code>
+      <code><![CDATA[queryStringsForEncoding]]></code>
+      <code><![CDATA[standardSchemePortCombinations]]></code>
+      <code><![CDATA[userInfoProvider]]></code>
+      <code><![CDATA[utf8PathsDataProvider]]></code>
+      <code><![CDATA[utf8QueryStringsDataProvider]]></code>
+      <code><![CDATA[validPorts]]></code>
+    </PossiblyUnusedMethod>
     <PossiblyUnusedParam>
       <code><![CDATA[$query]]></code>
     </PossiblyUnusedParam>
--- a/src/Uri.php
+++ b/src/Uri.php
@@ -10,6 +10,7 @@ use Stringable;
 
 use function array_keys;
 use function explode;
+use function filter_var;
 use function implode;
 use function ltrim;
 use function parse_url;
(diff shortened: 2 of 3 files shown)


Xerkus:
@laminas/technical-steering-committee this one is tricky. Needs feedback.

Interop integration tests do not establish how any of this should be treated. Other implementations allow any crap to be set as host and as a result produce junk url.

```
php > var_export(parse_url('http://(:1'));
array (
  'scheme' => 'http',
  'host' => '(',
  'port' => 1,
)
php > var_export(parse_url('http://,:1'));
array (
  'scheme' => 'http',
  'host' => ',',
  'port' => 1,
)
php > var_export(parse_url('http://[:1'));
array (
  'scheme' => 'http',
  'host' => '[',
  'port' => 1,
)
php > var_export(parse_url('http://]:1'));
array (
  'scheme' => 'http',
  'host' => ']',
  'port' => 1,
)
php > var_export(parse_url('http://][:1'));
array (
  'scheme' => 'http',
  'host' => '][',
  'port' => 1,
)
php > var_export(parse_url('http://;:1'));
array (
  'scheme' => 'http',
  'host' => ';',
  'port' => 1,
)
php > var_export(parse_url('http://::1'));
array (
  'scheme' => 'http',
  'host' => ':',
  'port' => 1,
)
php > var_export(parse_url('http:// :1'));
array (
  'scheme' => 'http',
  'host' => ' ',
  'port' => 1,
)
php > var_export(parse_url('http:// '));
array (
  'scheme' => 'http',
  'host' => ' ',
)
```

url_parse handles malformed IPv6 that is not enclosed in brackets but the last segment is treated as a port.
 
Malformed: `https://user:pass@fe80::200:5aee:feaa:20a2:3001/foo?bar=baz#quz`
Valid: `https://user:pass@[fe80::200:5aee:feaa:20a2]:3001/foo?bar=baz#quz`

Since this malformed format is accepted I feel we should allow to set IPv6 address as a host using `withHost()` but enclose it in brackets ourselves.
Alternative would be to reject IPv6 address that is not enclosed with brackets in all cases. It will be safer and frankly the proper approach but it also can be considered a breaking change.

For the remaining filtering introduced in this PR it only explicitly forbids general delimeters but allows anything else to run rampant.

I believe validation needs to be clamped down to only allow IPv6, IPv4 and valid for DNS hostnames. Probably pass it through validate domain filter. Any other URI variants are not suitable for http message context. Something along those lines:

```php
filter_var(idn_to_ascii($host), FILTER_VALIDATE_DOMAIN, FILTER_FLAG_HOSTNAME)
```

nyamsprod:
AFAIK the following 

```php
var_export(parse_url('http:// '));
```

should fail already because RFC3986 does not allow spaces in the URI the test should then be on
 
```php
var_export(parse_url('http://%20'));
```

But perhaps this is another issue which is not related to the current ticket.

PR #202, opened 2024-11-11
QA: Add type declarations and refine psalm types

ADmad:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

Some properties and methods were missing type declarations.

### DIFF
--- a/src/MessageTrait.php
+++ b/src/MessageTrait.php
@@ -30,16 +30,14 @@ trait MessageTrait
     /**
      * List of all registered headers, as key => array of values.
      *
-     * @var array
-     * @psalm-var array<non-empty-string, list<string>>
+     * @var array<non-empty-string, list<string>>
      */
     protected $headers = [];
 
     /**
      * Map of normalized header name to original name used to register header.
      *
-     * @var array
-     * @psalm-var array<non-empty-string, non-empty-string>
+     * @var array<non-empty-string, non-empty-string>
      */
     protected $headerNames = [];
 
--- a/src/Response/JsonResponse.php
+++ b/src/Response/JsonResponse.php
@@ -43,8 +43,7 @@ class JsonResponse extends Response
         | JSON_HEX_QUOT
         | JSON_UNESCAPED_SLASHES;
 
-    /** @var mixed */
-    private $payload;
+    private mixed $payload;
 
     /**
      * Create a JSON response with the given data.
--- a/src/Uri.php
+++ b/src/Uri.php
@@ -55,7 +55,11 @@ class Uri implements UriInterface, Stringable
      */
     public const CHAR_UNRESERVED = 'a-zA-Z0-9_\-\.~\pL';
 
-    /** @var int[] Array indexed by valid scheme names to their corresponding ports. */
+    /**
+     * Array indexed by valid scheme names to their corresponding ports.
+     *
+     * @var array<string, positive-int>
+     */
     protected $allowedSchemes = [
         'http'  => 80,
         'https' => 443,
--- a/src/Request/ArraySerializer.php
+++ b/src/Request/ArraySerializer.php
@@ -70,10 +70,9 @@ final class ArraySerializer
     }
 
     /**
-     * @return mixed
      * @throws Exception\DeserializationException
      */
-    private static function getValueFromKey(array $data, string $key, ?string $message = null)
+    private static function getValueFromKey(array $data, string $key, ?string $message = null): mixed
     {
         if (isset($data[$key])) {
             return $data[$key];
--- a/src/Response/ArraySerializer.php
+++ b/src/Response/ArraySerializer.php
@@ -68,10 +68,9 @@ final class ArraySerializer
     }
 
     /**
-     * @return mixed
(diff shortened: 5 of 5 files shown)


src/MessageTrait.php
```diff
@@ -30,24 +30,20 @@ trait MessageTrait
...
+    protected array $headers = [];
 
     /**
      * Map of normalized header name to original name used to register header.
      *
-     * @var array
      * @psalm-var array<non-empty-string, non-empty-string>
      */
-    protected $headerNames = [];
+    protected array $headerNames = [];
```
gsteel: Adding types to protected properties is a BC break, also, removing `@var array` is fine - it would be good to rename the `@psalm-var` annotations to `@var` - the psalm prefix is unnecessary

src/Response/JsonResponse.php
```diff
@@ -80,10 +79,7 @@ public function __construct(
         parent::__construct($body, $status, $headers);
     }
 
-    /**
-     * @return mixed
-     */
-    public function getPayload()
+    public function getPayload(): mixed
```
gsteel: Adding the return type is a BC break for inheritors: https://3v4l.org/BH4qV

src/Uri.php
```diff
@@ -53,7 +53,7 @@ class Uri implements UriInterface, Stringable
     public const CHAR_UNRESERVED = 'a-zA-Z0-9_\-\.~\pL';
 
     /** @var int[] Array indexed by valid scheme names to their corresponding ports. */
-    protected $allowedSchemes = [
+    protected array $allowedSchemes = [
```
gsteel: Adding types to protected members is a BC break: https://3v4l.org/frBrF

The documented type could be improved with `/** @var array<string, positive-int> */`

gsteel (changes requested):
Thanks for the patch @ADmad but there are a few BC breaks that need reverting

ADmad:
All these are private properties, shouldn't it be safe to change them?

```
[BC] CHANGED: Type of property Laminas\Diactoros\RequestTrait#$protocol changed from having no type to string
[BC] CHANGED: Type of property Laminas\Diactoros\RequestTrait#$stream changed from having no type to Psr\Http\Message\StreamInterface
[BC] CHANGED: Type of property Laminas\Diactoros\RequestTrait#$method changed from having no type to string
[BC] CHANGED: Type of property Laminas\Diactoros\RequestTrait#$requestTarget changed from having no type to string|null
[BC] CHANGED: Type of property Laminas\Diactoros\RequestTrait#$uri changed from having no type to Psr\Http\Message\UriInterface
[BC] CHANGED: Type of property Laminas\Diactoros\MessageTrait#$protocol changed from having no type to string
[BC] CHANGED: Type of property Laminas\Diactoros\MessageTrait#$stream changed from having no type to Psr\Http\Message\StreamInterface
```

Constructors can be changed with impunity, changing the type shouldn't cause any error even if the constructor was overridden in a child class. 

```
[BC] CHANGED: The parameter $data of Laminas\Diactoros\Response\JsonResponse#__construct() changed from no type to mixed
```

gsteel:
The change in JsonResponse constructor is fine IMO, but it _is_ a technical BC break, just because PHP lets us change the signature of __construct, it's still part of the public API.

The private property changes to the traits are indeed a BC break if users have re-implemented a property without a type hint: https://3v4l.org/aaQTo - if they were class properties, it'd be fine, so very little of this can go into a 3.x release

ADmad:
Thank you for the trait example, I'll revert the related changes.

I am still not convinced the constructor change is a BC break :)
The arguments aren't being changed, just the untyped argument is explicitly typed as `mixed`. I can't think of any practical use case which would generate an error with this change.

ADmad:
I don't think the test failures are related to the changes in this PR.

gsteel:
No, but whatever they are, they'll need to be addressed separately 🤔

gsteel:
See #205

ADmad:
That's not a "me" problem, so I'll let you guys tackle that :)

ADmad:
Ping

froschdesign:
@ADmad 
> Ping

Please could you explain what you expect?

ADmad:
@froschdesign I expect the PR to be merged :) I believe I have made the changes as per the feedback. The test failures are unrelated to my changes.

froschdesign:
@ADmad 
> The test failures are unrelated to my changes.

Thanks for you response but a fix for the problem is needed before something can merge here. And if another pull request is required beforehand, this one remains open.
Merging this pull request means ready for release, but it is not, because there are still open problems.
Thanks for your understanding!

ADmad:
Okay, thank you for the explanation.

ADmad:
With the base branch for the merge being updated CI is green now.

Xerkus:
Please do a rebase instead of merge to update

ADmad:
> Please do a rebase instead of merge to update

Done

Xerkus (approved):

PR #214, opened 2025-08-11
add PHP 8.5 and remove 8.1

### LINKED ISSUE #218: Add support for PHP 8.5

LordSimal:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

[PHP 8.5 is on its way](https://wiki.php.net/todo/php85) so any library using this as a base should be able to `composer install` with this package as part of their dependency

We - as the CakePHP Core Team - currently are not able to test PHP 8.5 due to this version constraint missing.

I'd personally would change that to `^8.1` but I guess you selectively wan't to support new minor versions.


froschdesign:
Without PHP 8.5 in the CI pipeline, this cannot be checked and merged. We have to wait until a suitable pre-release version of PHP can be integrated into the CI pipeline.
Anything else would mean we are running across the street with our eyes closed.

LordSimal:
Thanks for the quick response. totally understandable, I will be patient 👍🏻 😉

froschdesign:
Thank you for your understanding! 👍 
Unfortunately, PHP is [not backward compatible](https://www.php.net/releases/8.4/en.php#deprecations_and_bc_breaks), and in order to ensure quality, it must first be tested with the appropriate version.

arshidkv12:
I ran PHPUnit tests with PHP 8.5.0beta1 and didn’t notice any issues. https://php.watch/versions/8.5/releases/8.5.0beta1

LordSimal:
RC1 is here, so how about now? 😁 
https://www.php.net/archive/2025.php#2025-09-25-3

gsteel:
Nothing will happen until 8.5 is available in Debian packages. After that, CI containers will need to be updated before laminas & mezzio libs can be tested under 8.5. Support for 8.5 will only be merged and released once CI is passing

Ocramius:
Given #209, this should now be rebased

LordSimal:
So now CI is failing because of https://github.com/vimeo/psalm/pull/11541

Xerkus:
Also drop PHP 8.1 at the same time. We would not be waiting until end of december.

Here is an example of the changes you would need to do to make it work https://github.com/mezzio/mezzio-session/pull/78
Except for phpunit changes as 11 is currently not installable due to another dependency preventing it https://github.com/php-http/psr7-integration-tests/pull/80

Ocramius:
`composer.lock` needs refreshing too, so does `config -> platform -> php` in `composer.json`

LordSimal:
Sorry, didn't know that. I just did a composer update -W with PHP 8.2, is that correct?

Ocramius:
@LordSimal no worries, can't know what to do without it being somehow learned somewhere :-) 

Anyway, we'll need to tell PHP 8.5 CI to ignore dependency checks for the time being.

See https://github.com/laminas/laminas-stdlib/blob/b1c81514cfe158aadf724c42b34d3d0a8164c096/.laminas-ci.json#L2-L3 for an example

Xerkus (changes requested):
```
PHP Deprecated:  Function curl_close() is deprecated since 8.5, as it has no effect since PHP 8.0 in /github/workspace/test/ResponseTest.php on line 112
```

@LordSimal please drop the function from the test and we are good to go.

Xerkus (approved):

Xerkus:
@LordSimal thank you.

PR #216, opened 2025-08-17
Fix “Unsafe usage of new static()” by making EmptyResponse final

arshidkv12:
Marking EmptyResponse as final prevents late static binding issues
when instantiating with `new static()`, ensuring type safety and
avoiding potential runtime errors in subclasses.

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | yes
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Marking EmptyResponse as final prevents late static binding issues
when instantiating with `new static()`, ensuring type safety and
avoiding potential runtime errors in subclasses.

### DIFF
--- a/src/Response/EmptyResponse.php
+++ b/src/Response/EmptyResponse.php
@@ -10,7 +10,7 @@ use Laminas\Diactoros\Stream;
 /**
  * A class representing empty HTTP responses.
  */
-class EmptyResponse extends Response
+final class EmptyResponse extends Response
 {
     /**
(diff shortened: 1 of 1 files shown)


samsonasik:
This is bc break, can't land in minor version.

There should be no child for this class, extends parent Response instead.

If needed, use
 ```
/** @final */
```

so no bc break, while inform user to not extends it.

PR #220, opened 2025-10-12
Allow PHP 8.1

TimWolla:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | not really
| RFC           | no
| QA            | no

### Description

PHP 8.1 will not be EOL until December 31st and the code still is trivially compatible with PHP 8.1. Re-add support for PHP 8.1 for 3.8.x to give users one branch with additional flexibility to support a wide(r) range of PHP versions.


Xerkus:
Support for 8.1 is dropped at the same time as we add support for 8.5
 
https://github.com/laminas/technical-steering-committee/blob/main/meetings/minutes/2025-10-06-TSC-Minutes.md#adding-85-support-everywhere

> The TSC members decided in an asyncronous poll that once PHP 8.5 support will be released in a package, the support for 8.1 will be removed in the same PR. In order to have only 4 PHP versions in composer.json, and reduce the release burden.

8.1 support is available in the previous version if you need it.

PR #230, opened 2026-07-19
Allow HTTP/3 protocol versions

masakielastic:
## Description

Update protocol version validation to accept HTTP/3 protocol versions.

`MessageTrait::validateProtocolVersion()` currently accepts the following values:

* `1.0`
* `1.1`
* `2`
* `2.0`

However, passing `3` to `withProtocolVersion()` results in an `InvalidArgumentException`.

This change adds support for both `3` and `3.0`, following the existing handling of HTTP/2.

### DIFF
--- a/test/MessageTraitTest.php
+++ b/test/MessageTraitTest.php
@@ -46,6 +46,8 @@ final class MessageTraitTest extends TestCase
             '1-without-minor'      => ['1'],
             '1-with-invalid-minor' => ['1.2'],
             '1-with-hotfix'        => ['1.1.1'],
+            '3-with-invalid-minor' => ['3.1'],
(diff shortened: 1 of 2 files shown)


froschdesign:
@weierophinney @Xerkus 
Can you spot any problems?

test/MessageTraitTest.php
```diff
@@ -65,6 +67,7 @@ public static function validProtocolVersionProvider(): array
             '1.1' => ['1.1'],
             '2'   => ['2'],
             '2.0' => ['2.0'],
+            '3.0' => ['3.0'],
```
weierophinney, line 70: Please add a test for "3" as well. But with that in place, I'm 👍

weierophinney (changes requested):

